PR #205, opened 2019-07-22
Add integration with assemble numbers

ben-pr-p:
Partially hidden API key:
Screen Shot 2019-07-22 at 3 59 50 PM

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -671,7 +671,9 @@ class AdminCampaignEdit extends React.Component {
           if (sectionIsSaving) {
             avatar = <CircularProgress style={avatarStyle} size={25} />;
             cardHeaderStyle.background = theme.colors.lightGray;
-            cardHeaderStyle.width = `${savePercent}%`;
+            cardHeaderStyle.width = `${
+              savePercent > 100 ? savePercent - 100 : savePercent
+            }%`;
           } else if (sectionIsExpanded && sectionCanExpandOrCollapse) {
             cardHeaderStyle.backgroundColor = theme.colors.lightYellow;
           } else if (!sectionCanExpandOrCollapse) {
@@ -725,7 +727,9 @@ class AdminCampaignEdit extends React.Component {
               </CardText>
               {sectionIsSaving && adminPerms ? (
                 <CardActions>
-                  <div>Current Status: {savePercent}% complete</div>
+                  <div>
+                    Current Status: {extractStageAndStatus(savePercent)}
+                  </div>
                   {jobMessage ? <div>Message: {jobMessage}</div> : null}
                   <RaisedButton
                     label="Discard Job"
@@ -739,6 +743,14 @@ class AdminCampaignEdit extends React.Component {
         })}
       </div>
     );
+  }
+}
+
+function extractStageAndStatus(percentComplete) {
+  if (percentComplete > 100) {
+    return `Filtering out landlines. ${percentComplete - 100}% complete`;
+  } else {
+    return `Uploading. ${percentComplete}% complete`;
   }
 }
 
--- a/src/containers/Settings.jsx
+++ b/src/containers/Settings.jsx
@@ -381,6 +381,7 @@ class Settings extends React.Component {
         <Card className={css(styles.sectionCard)}>
           <GSForm
             schema={numbersApiKeySchema}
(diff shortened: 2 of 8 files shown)


ben-pr-p:
Progress bar: 
Screen Shot 2019-07-23 at 1 17 37 PM

and 


Screen Shot 2019-07-23 at 1 17 55 PM

src/containers/AdminCampaignEdit.jsx
```diff
@@ -742,6 +746,14 @@ class AdminCampaignEdit extends React.Component {
   }
 }
 
+function extractStageAndStatus(percentComplete) {
```
bchrobot: This should take the name of the section as well. Non-contact sections should still say
```
`Current status: ${savePercent}% complete`
```

src/server/api/schema.js
```diff
@@ -150,9 +150,7 @@ const replaceShortLinkDomains = async (organizationId, messageText) => {
 
   const replacerReducer = (text, domain) => {
     const domainRegex = RegExp(
-      "/(https?:\/\/)"
-      + escapeRegExp(domain)
-      + "(:*)/g"
+      "/(https?://)" + escapeRegExp(domain) + "(:*)/g"
```
bchrobot: Necessary for this feature?

Also, are we sure the slashes don't need to be escaped? I pulled this from a Stack Overflow post (bad on me for not adding a comment with a link to the post) so I don't know.

src/server/api/schema.js
```diff
@@ -2291,6 +2289,23 @@ const rootMutations = {
...
+    setNumbersApiKey: async (
+      _,
+      { organizationId, numbersApiKey },
+      { user }
+    ) => {
+      await accessRequired(user, organizationId, "OWNER");
+
+      const organization = await Organization.get(organizationId);
+      const featuresJSON = JSON.parse(organization.features || "{}");
+      featuresJSON.numbersApiKey = numbersApiKey;
```
bchrobot: Check to make sure they haven't hit "Save" on the obfuscated field.

```suggestion
      if (featuresJSON.numbersApiKey === numbersApiKey.slice(0, 4) + "****************") return organization;
      featuresJSON.numbersApiKey = numbersApiKey;
```

src/containers/Settings.jsx
```diff
@@ -373,6 +377,28 @@ class Settings extends React.Component {
             </CardActions>
           </GSForm>
         </Card>
+
+        <Card className={css(styles.sectionCard)}>
+          <GSForm
+            schema={numbersApiKeySchema}
+            onSubmit={this.props.mutations.setNumbersApiKey}
```
bchrobot: Prevent user from saving without a change to the value (to prevent saving the obfuscated value). This needs a change to the state definition as well.

```suggestion
            onChange={({ numbersApiKey: newValue }) => this.setState({ hasNumbersApiKeyChanged: newValue !== numbersApiKey })}
            onSubmit={this.props.mutations.setNumbersApiKey}
```

src/containers/Settings.jsx
```diff
@@ -373,6 +377,28 @@ class Settings extends React.Component {
...
+              To enable automatic filtering of landline phone numbers, you will
+              need to put in your Assemble Numbers API Key here.
+              <Form.Field
+                label="Assemble Numbers API Key"
+                name="numbersApiKey"
+                fullWidth
+              />
+            </CardText>
+            <CardActions>
+              <Form.Button type="submit" label={"Save"} />
```
bchrobot:
```suggestion
              <Form.Button
                type="submit"
                label={"Save"}
                disabled={!this.state.hasNumbersApiKeyChanged}
              />
```

src/server/api/schema.js
ben-pr-p: This was a prettier change – I think slashes only need to be escaped if they are inside of slashes (standard regex definition)

ben-pr-p: Tested in Node shell that prettier is fine here and they don't need to be escaped when inside of quotes

src/server/api/schema.js
ben-pr-p: wow great catch!

ben-pr-p: i'll add a check for no `*`s, since those won't be in the key

joemcl:
thanks @ben-pr-p for pointing me here.

PR #226, opened 2019-07-30
fix: unassign on escalate

ben-pr-p:


src/server/api/schema.js
```diff
@@ -1901,6 +1915,8 @@ const rootMutations = {
         contactsFilter
       );
 
+      console.log(campaignContactIdsToMessageIds);
```
bchrobot:
```suggestion
```

src/server/api/schema.js
```diff
@@ -1901,6 +1915,8 @@ const rootMutations = {
...
+
```
bchrobot:
```suggestion
```

src/server/api/schema.js
```diff
@@ -1397,7 +1397,7 @@ const rootMutations = {
       const [deleteResult, insertResult] = await Promise.all([
         await r
           .knex("campaign_contact_tag")
-          .where({ campaign_contact_id: campaignContactId })
+          .where({ campaign_contact_id: parseInt(campaignContactId) })
```
bchrobot: Ah, good catch. This was working just passing the string, but better to be explicit.

PR #244, opened 2019-08-07
Make landline filtering optional on a per campaign basis

ben-pr-p:


src/containers/AdminCampaignEdit.jsx
```diff
@@ -808,6 +812,7 @@ const mapQueriesToProps = ({ ownProps }) => ({
             lastName
             displayName
           }
+          numbersApiKey
```
bchrobot: This is returned partially hidden, right?

src/containers/Settings.jsx
```diff
@@ -358,11 +365,17 @@ class Settings extends React.Component {
...
+                numbersApiKey: newValue
               })
             }
-            onSubmit={this.props.mutations.setNumbersApiKey}
-            defaultValue={{ numbersApiKey }}
+            onSubmit={this.doSetNumbersApiKey}
+            defaultValue={{
+              numbersApiKey:
+                this.state.numbersApiKey === undefined
+                  ? numbersApiKey
```
bchrobot: Set the initial state based on props.data.organization and then make `GSForm` controlled?

src/containers/AdminCampaignEdit.jsx
bchrobot: And further down in the PR my question is answered.

PR #251, opened 2019-08-16
feat: gcs support

bchrobot:


src/workers/exports/gs.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Get a signed URL for an object that is valid for 24 hours.
+ *
+ * @param {string} bucket Name of the S3 bucket to upload the object to
+ * @param {string} key Name of key of destination object
+ * @returns {string} Signed download URL
+ */
+const getDownloadUrl = async (bucket, key) => {
+  throw new Error("Unimplemented!");
```
ben-pr-p: What's up here?

bchrobot: ah darn, I messed up the PR description. This is getting punted because it's gonna be some more work to get it right. We only need `gs-json` and `s3` to support our current GKE and AWS deploys

PR #254, opened 2019-08-19
feat: logging

bchrobot:
* Lots of find-and-replace.
* Removed the wrap function
* Respond with error codes when message handling fails (beyond scope, I know 😞)


src/lib/phone-format.js
```diff
@@ -1,5 +1,5 @@
 import { PhoneNumberUtil, PhoneNumberFormat } from "google-libphonenumber";
-import { log } from "./log";
+import logger from "../logger";
```
ben-pr-p: Doesn't this file get included on the client too?

src/lib/zip-format.js
```diff
@@ -98,7 +100,7 @@ const findZipRanges = function(r) {
       zipchanges.sort(function(a, b) {
         return b[4] - a[4];
       });
-      console.log(zipchanges);
+      logger.info(zipchanges);
```
ben-pr-p: I think this one might as well

src/lib/phone-format.js
bchrobot: There was one unused reference to `getDisplayPhoneNumber` that has been removed in 5ee63c3

src/lib/zip-format.js
bchrobot: There are no client references to this!

PR #263, opened 2019-08-22
feat(immutable-question-response): migration

ben-pr-p:
This migration implements immutable question responses
by renaming question_response to all_question_response,
creating a view named question_response, and binding
triggers such that queries against question_response
such that updates, inserts, and deletes at the question_response
view leave a record of modification in the all_question_response
table.

Importantly, the migration can be run and rolled back without any
loss of fresh data, but rolling back a migration will delete
the historical records of previous question responses, since
it deletes the columns that allow one to distinguish between an old
and new question response.

In the future, if the number of old question responses begins to
slow down queries, modifying indexes on question_response to
be partial – only on is_deleted = false – will speed them up.

### DIFF
--- a/migrations/20190822080351_immutable_quesiton_responses.js
+++ b/migrations/20190822080351_immutable_quesiton_responses.js
@@ -1,0 +1,136 @@
+/**
+ * Invalidate overlapping quesiton responses on insert
+ *
+ * Runs before insert so that it does not modify the NEW row
+ * Only rows that the NEW row is replacing
+ * Return NEW to allow the insert to proceed unmodified
+ *
+ * This trigger is on the 'all_question_response' table because
+ * views cannot have before/after insert triggers
+ *
+ * Since the view is simple (involving no joins), it's insert, updates, and deletes
+ * are just forwarded
+ *
+ * The extra "is_deleted = false" is not necessary for proper correctness, only
+ * to ensure that the updated_at is not reset and a more accurate history is preserved
+ */
+const BEFORE_INSERT_FUNCTION = `
+create or replace function all_question_response_before_insert() returns trigger as $$
+begin
+  update all_question_response
+  set is_deleted = true, updated_at = now()
+  where campaign_contact_id = NEW.campaign_contact_id
+    and interaction_step_id = NEW.interaction_step_id
+    and is_deleted = false;
+
+  return NEW;
+end;
+$$ language plpgsql;
+`;
+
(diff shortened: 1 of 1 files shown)


migrations/20190822080351_immutable_quesiton_responses.js
```diff
@@ -0,0 +1,136 @@
...
+ * The new record will have an updated updated_at
+ * Return NEW to allow the update to proceed unmodified
+ *
+ * Since the view is simple (involving no joins), it's insert, updates, and deletes
+ * are just forwarded
+ *
+ * This is only bound to conventional updates – not update set is_deleted = true
+ */
+const BEFORE_UPDATE_FUNCTION = `
+create or replace function all_question_response_before_update() returns trigger as $$
```
bchrobot: This will result in non-sequential `id`s. That's fine, but may give pause to anyone who is expecting sequential question responses to have sequential `id`s.

Instead of copying `OLD` to a new record, what if we insert `NEW` as the new record and then return `OLD` unmodified except for `is_deleted` and `updated_at`?

migrations/20190822080351_immutable_quesiton_responses.js
```diff
@@ -0,0 +1,136 @@
...
+ * Must return OLD for command signature / returning statements to function correctly
+ *
+ * This is an instead of trigger, so it can be bound directly to the view
+ */
+const INSTEAD_OF_DELETE_FUNCTION = `
+create or replace function question_response_instead_of_delete() returns trigger as $$
+begin
+  update all_question_response
+  set is_deleted = true, updated_at = now()
+  where id = OLD.id;
```
bchrobot: Worth making this defensive?

```suggestion
  where campaign_contact_id = OLD.campaign_contact_id
  and interaction_step_id = OLD.interaction_step_id;
```

ben-pr-p: This would break simultaneous insert and delete by soft deleting the just inserted record!

migrations/20190822080351_immutable_quesiton_responses.js
ben-pr-p: Yeah – I see the problem. I feel if `update question_response set some_column_that_isnt_id = some_value` changes the primary key behind the scenes, that violates principle of least surprise if one is interacting with the view as if it were a table though, which is the desired model

PR #290, opened 2019-09-18
feat(monitoring): add healthcheck endpoint

bchrobot:
Currently, Datadog network availability monitors do not catch application errors due to database inaccessibility as `/` does not require any database queries if there is no user session.


src/server/routes/utils.js
```diff
@@ -31,4 +31,15 @@ router.post("/remove-number-from-campaign", async (req, res) => {
   return res.sendStatus(200);
 });
 
+// The health check endpoint should ensure that the database is reachable
+router.get("/health", async (req, res) => {
+  try {
+    await r.knex.raw("select sd1;");
```
ben-pr-p:
```suggestion
    await r.knex.raw("select 1;");
```

PR #291, opened 2019-09-20
feat(sweeping): add previous/next conversation navigation

bchrobot:
This allows a sweeper to navigate to the previous/next conversation without closing the current
conversation and opening the next one. It will only go forward/backward on the current page of
results, however; navigating to the next page of results manually is still required.

Spoke

### DIFF
--- a/src/components/IncomingMessageList/index.jsx
+++ b/src/components/IncomingMessageList/index.jsx
@@ -48,7 +48,7 @@ function prepareSelectedRowsData(conversations, rowsSelected) {
 
 export class IncomingMessageList extends Component {
   state = {
-    activeConversation: undefined
+    activeConversationIndex: -1
   };
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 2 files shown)


src/components/IncomingMessageList/index.jsx
```diff
@@ -212,17 +212,27 @@ export class IncomingMessageList extends Component {
...
   };
 
-  handleCloseConversation = () => {
-    this.setState({ activeConversation: undefined });
+  handleRequestNextConversation = () => {
+    const { activeConversationIndex: oldIndex } = this.state;
+    const { conversations } = this.props.conversations.conversations;
+    const newIndex = oldIndex + 1;
+    if (newIndex >= conversations.length) return;
+    this.setState({ activeConversationIndex: newIndex });
```
ben-pr-p: Idiomatic way to do this is:

```suggestion
    this.setState(prevState => {
      const { activeConverastionIndex: oldIndex } = this.state;
      const { conversations } = this.props.conversations.conversations;
      const newIndex = oldIndex + 1;
      if (newIndex >= conversations.length) return;
      return { activeConversationIndex: newIndex };
    });
```

That way if someone spams the button, it'll increment – otherwise, because `this.setState` is async, the next `handleRequestNextConversation` may have the old state

PR #293, opened 2019-09-20
fix: campaign list size

bchrobot:


src/server/api/campaign.js
```diff
@@ -57,7 +57,10 @@ export async function getCampaigns(organizationId, cursor, campaignsFilter) {
   campaignsQuery = campaignsQuery.orderBy("due_by", "desc").orderBy("id");
 
   if (cursor) {
-    campaignsQuery = campaignsQuery.limit(cursor.limit).offset(cursor.offset);
+    // A limit of 0 means a page size of 'All'
```
ben-pr-p: Only ok because it's Spoke

PR #302, opened 2019-09-26
feat: contact level time zone assignment

ben-pr-p:


migrations/20190926104147_contact-level-timezone-assignment.js
```diff
@@ -0,0 +1,143 @@
...
+     * a view if its columns change, and assignable_campaign_contacts adds a contact_timezone column
+     *
+     * assignable_needs_message and assignable_needs_reply must select their columns directly
+     * to avoid requiring a drop and recreate (can't add new columns)
+     *
+     * finally, once it's safe to do so, we remove the time zone constraint from the campaign level view
+     */
+    `
+    create or replace function spoke_tz_to_iso_tz(spoke_tz text) returns text as $$
+      select case
```
bchrobot: Are there not `_0`, `_1` pairs for each hour offset?

migrations/20190926104147_contact-level-timezone-assignment.js
```diff
@@ -0,0 +1,143 @@
...
+        )
+    );
+
+    create or replace view assignable_needs_message as (
+      select acc.id, acc.campaign_id, acc.message_status
+      from assignable_campaign_contacts as acc
+      where message_status = 'needsMessage'
+        and (
+          acc.texting_hours_end > extract(hour from (CURRENT_TIMESTAMP at time zone acc.contact_timezone) + interval '10 minutes')
+          or ( acc.contact_timezone = 'UNKNOWN' and extract(hour from CURRENT_TIMESTAMP) < 21 and extract(hour from CURRENT_TIMESTAMP) > 12 )
```
bchrobot: Comment that we fall back to most conservative timezone range for continental US

migrations/20190926104147_contact-level-timezone-assignment.js
```diff
@@ -0,0 +1,143 @@
...
+            and campaign_contact_tag.campaign_contact_id = campaign_contact.id
+        )
+    );
+
+    create or replace view assignable_needs_message as (
+      select acc.id, acc.campaign_id, acc.message_status
+      from assignable_campaign_contacts as acc
+      where message_status = 'needsMessage'
+        and (
+          acc.texting_hours_end > extract(hour from (CURRENT_TIMESTAMP at time zone acc.contact_timezone) + interval '10 minutes')
```
bchrobot: How does this handle `acc.contact_timezone = 'UNKNOWN'`?

migrations/20190926104147_contact-level-timezone-assignment.js
```diff
@@ -0,0 +1,143 @@
...
+};
+
+exports.down = function(knex, Promise) {
+  return knex.schema.raw(
+    /**
+     * Bring back campaign restriction
+     *
+     * Remove contact restriction
+     *
+     * No need to destroy function / remove column from assignable_contacts
```
bchrobot: `down` should completely undo `up`

migrations/20190926104147_contact-level-timezone-assignment.js
ben-pr-p: Nope, did a distinct

ben-pr-p:
> What is the point of `old_assignable_campaign_contacts` and the first definitions of `assignable_needs_message` (46) and `assignable_needs_reply` (52)? They are subsequently either dropped or redefined without an intermediate reference.

See https://github.com/politics-rewired/Spoke/pull/302/files#diff-f0c32a69c67a5cf1e570880226405c85R10

migrations/20190926104147_contact-level-timezone-assignment.js
ben-pr-p: It was erroring – with the order switched, it early returns

migrations/20190926104147_contact-level-timezone-assignment.js
ben-pr-p: I don't think so – down should restore the state of the application to the way it was

bchrobot: Right, but the way that is was did not include the `spoke_tz_to_iso_tz` function. So doing up, down, up, down leaves behind `spoke_tz_to_iso_tz`. I think it's much better to be precise than to have to worry about the potential problems resulting from side effects like this compounding over time.

@ben-pr-p

PR #311, opened 2019-10-11
feat: Assemble Numbers SMS support

bchrobot:
There are a few TODOs still

### DIFF
--- a/src/server/api/lib/assemble-numbers.js
+++ b/src/server/api/lib/assemble-numbers.js
@@ -1,0 +1,314 @@
+import NumbersClient from "assemble-numbers-client";
+
+import { config } from "../../../config";
+import logger from "../../../logger";
+import { r } from "../../models";
+import { getFormattedPhoneNumber } from "../../../lib/phone-format";
+import { sleep } from "../../../lib/utils";
+import { symmetricDecrypt } from "./crypto";
+import {
+  SpokeSendStatus,
+  getMessagingServiceById,
+  getContactMessagingService,
+  appendServiceResponse,
+  messageComponents,
+  getCampaignContactAndAssignmentForIncomingMessage,
+  saveNewIncomingMessage
+} from "./message-sending";
+
+export const NumbersSendStatus = Object.freeze({
+  Queued: "queued",
+  Sending: "sending",
+  Sent: "sent",
+  Delivered: "delivered",
+  SendingFailed: "sending_failed",
+  DeliveryFailed: "delivery_failed",
+  DeliveryUnconfirmed: "delivery_unconfirmed"
+});
+
+/**
+ * Create an Assemble Numbers client
+ * @param {object} numbersService Assemble Numbers Messaging Service record
+ * @returns Assemble Numbers JS client
+ */
+export const numbersClient = async numbersService => {
+  const encryptedApiKey = numbersService.encrypted_auth_token;
+  const apiKey = symmetricDecrypt(encryptedApiKey);
+  const client = new NumbersClient({ apiKey });
+  return client;
+};
+
+export const inboundMessageValidator = () => async (req, res, next) => {
+  // Check if the 'x-assemble-signature' header exists or not
+  if (!req.header("x-assemble-signature")) {
+    return res
+      .type("text/plain")
+      .status(400)
+      .send(
+        "No signature header error - x-assemble-signature header does not exist, maybe this request is not coming from Assemble."
+      );
+  }
+
+  const signature = req.header("x-assemble-signature");
+  const { id: messageId, profileId } = req.body;
+
+  const service = await getMessagingServiceById(profileId);
+  const numbers = await numbersClient(service);
+  const isValid = numbers.validateInboundMessageWebhook(messageId, signature);
+
+  if (isValid) {
+    next();
+  } else {
+    return res
+      .type("text/plain")
+      .status(403)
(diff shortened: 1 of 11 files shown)


src/server/api/lib/message-sending.js
```diff
@@ -1,4 +1,222 @@
...
+  const serviceSid = await assignMessagingServiceSID(cell, organization_id);
+  const messagingService = await r
+    .knex("messaging_service")
+    .where({ messaging_service_sid: serviceSid })
+    .first();
+  return messagingService;
+};
+
+/**
+ * Make best effort attempty to assign messaging services to all campaign contacts in a campaign
```
ben-pr-p:
```suggestion
 * Make best effort attempt to assign messaging services to all campaign contacts in a campaign
```

src/server/api/lib/message-sending.js
```diff
@@ -1,4 +1,222 @@
...
+    .where({ messaging_service_sid: messagingServiceId })
+    .first();
+
+/**
+ * Fetches an existing assigned messaging service for a campaign contact. If no messaging service
+ * has been assigned then assign one and return that.
+ * @param {number} campaignContactId The ID of the target campaign contact
+ * @returns {object} Assigned messaging service Postgres row
+ */
+export const getContactMessagingService = async campaignContactId => {
```
ben-pr-p:
```suggestion
export const getContactMessagingService = async (campaignContactId, organizationId) => {
```

This could get passed the organizationId from [here](https://github.com/politics-rewired/Spoke/pull/311/files#diff-5d117a0afdd02753cbd75a8890706ee4R168), saving 1 round trip to DB

src/server/api/lib/message-sending.js
```diff
@@ -1,4 +1,222 @@
...
+  // Return an existing match if there is one
+  if (existingMessagingService) return existingMessagingService;
+
+  // Otherwise select an appropriate messaging service and assign
+  const serviceSid = await assignMessagingServiceSID(cell, organization_id);
+  const messagingService = await r
+    .knex("messaging_service")
+    .where({ messaging_service_sid: serviceSid })
+    .first();
+  return messagingService;
```
ben-pr-p: Looks like `assignMessagingServiceSID` is only used here – let's make it return everything we need so we don't have to do another query?

src/server/api/lib/message-sending.js
```diff
@@ -1,4 +1,222 @@
...
+ * @returns {string} The (S)ID of the messaging service assigned to that (cell, organization)
+ */
+export const assignMessagingServiceSID = async (cell, organizationId) => {
+  const {
+    rows: [{ messaging_service_sid }]
+  } = await r.knex.raw(
+    `
+      with chosen_messaging_service_sid as (
+        select
+          messaging_service.messaging_service_sid,
```
ben-pr-p:
```suggestion
          messaging_service.*,
```

return everything to avoid another round trip!

src/server/api/lib/assemble-numbers.js
```diff
@@ -0,0 +1,314 @@
...
+    mediaUrls,
+    contactZipCode
+  };
+  try {
+    await r.knex.transaction(async trx => {
+      let { service_response } = await trx("message")
+        .where({ id: spokeMessageId })
+        .first("service_response");
+      const result = await numbers.sms.sendMessage(messageInput);
+      console.log("raw result", result);
```
ben-pr-p:
```suggestion
```

Remove `console.log`

src/server/api/lib/assemble-numbers.js
```diff
@@ -0,0 +1,314 @@
...
+        .where({ id: spokeMessageId });
+    });
+  } catch (exc) {
+    console.error("Error sending message with Assemble Numbers", exc);
+    await r
+      .knex("message")
+      .update({
+        send_status: SpokeSendStatus.Error,
+        sent_at: r.knex.fn.now(),
+        service_response
```
ben-pr-p:
```suggestion
        service_response: appendServiceResponse(service_response, exc)
```

src/server/api/lib/assemble-numbers.js
```diff
@@ -0,0 +1,314 @@
...
+      // TODO: assemble-numbers-client squashes errors
+      if (!sendMessage) throw new Error("Missing sendMessage in payload!");
+
+      service_response = appendServiceResponse(service_response, result);
+      await r
+        .knex("message")
+        .update({
+          send_status: SpokeSendStatus.Sent,
+          sent_at: r.knex.fn.now(),
+          service_response
```
ben-pr-p:
```suggestion
          service_response: JSON.stringify([result])
```

src/server/api/lib/assemble-numbers.js
ben-pr-p: Actually this should be a single error with just the error

src/server/api/lib/message-sending.js
bchrobot: The goal was to write a general purpose method here that didn't require passing (and first fetching) redundant information. The first query can't be completely removed as we still have to fetch the contact's cell.

It could be moved into the second query as a CTE, but then you lose the validation that the contact actually exists. That is probably fine, however.

ben-pr-p: It makes sense to fetch here because you need to switch on it for what service you use to send, right? But organization ID will always have been already queried in this request lifecycle

PR #313, opened 2019-10-11
fix: messaging service assignment

bchrobot:


src/server/api/lib/message-sending.js
```diff
@@ -134,8 +137,12 @@ export const getContactMessagingService = async campaignContactId => {
     ...existingMessagingService
   } = lookupResult;
 
+  const isRealService =
+    Object.values(existingMessagingService).filter(val => val !== null)
+      .length === 6;
```
ben-pr-p: Let's check for nullness of `existingMessagingService.messaging_service_sid` instead

PR #314, opened 2019-10-15
fix: include unmessaged contacts in message review queries

ben-pr-p:


src/server/api/conversations.js
```diff
@@ -121,7 +121,7 @@ export async function getConversations(
     contactNameFilter
   )).query;
 
-  offsetLimitQuery = offsetLimitQuery.join(
+  offsetLimitQuery = offsetLimitQuery.leftJoin(
```
bchrobot: `.orderBy("message.created_at", "desc")` sorts null values to the top. Should this have a primary sort on `COALESCE (message.created_at, campaign_contact.updated_at)` instead?

ben-pr-p: Probably!

ben-pr-p: https://github.com/politics-rewired/Spoke/pull/314/commits/01ae8fa1a75e7795a66c3df00c8f8548e3cf39d3

PR #316, opened 2019-10-16
fix: restrict contact upload to 75% of db pool size

bchrobot:


src/workers/jobs.js
```diff
@@ -33,7 +33,7 @@ import gsJson from "./exports/gs-json";
 import zipCodeToTimeZone from "zipcode-to-timezone";
 
 const CHUNK_SIZE = 1000;
-const BATCH_SIZE = config.DB_MAX_POOL;
+const BATCH_SIZE = Math.max(1, Math.floor(config.DB_MAX_POOL * 0.75));
```
ben-pr-p:
```suggestion
const BATCH_SIZE = Math.max(1, Math.floor(config.DB_MAX_POOL * 0.5));
```

PR #326, opened 2019-10-18
feat: search by unassigned (null texter)

bchrobot:


src/server/api/conversations.js
```diff
@@ -3,6 +3,7 @@ import { config } from "../../config";
 import { Assignment, r } from "../models";
 import { addWhereClauseForContactsFilterMessageStatusIrrespectiveOfPastDue } from "./assignment";
 import { buildCampaignQuery } from "./campaign";
+import { UNASSIGNED_TEXTER } from "../../components/IncomingMessageFilter";
```
ben-pr-p: I think, given the previous things build issues we've dealt with, it's best not to do server side imports from jsx. I could be persuaded if there were substantial benefit, but it's easy enough to just reproduce this constant either in this file or on a server side constants file

bchrobot: True -- how about in a top level .`/src/lib/constants.js` that has no dependencies itself?

ben-pr-p: That's fine! I think defining it in this file is fine too, I don't think there's benefits to extracting it if the constants aren't shared across multiple files

bchrobot: Sorry, I meant a top level constants file that could be shared across client and server. Having no dependencies of its own would mean it wouldn't introduce any client<>server dependency issues

ben-pr-p: Oh sure yeah!

PR #336, opened 2019-10-25
fix: let delivered override error

ben-pr-p:


src/server/api/lib/assemble-numbers.js
```diff
@@ -183,20 +183,24 @@ export const handleDeliveryReport = async reportBody => {
...
         .where({ service_id: messageId })
-        // Ignore late-arrival status updates for messages already in a final state
-        .whereNotIn("send_status", [
-          SpokeSendStatus.Delivered,
-          SpokeSendStatus.Error
-        ])
-    )
+        .first("send_status");
+
+      if (message.send_status !== SpokeSendStatus.Delivered) {
```
bchrobot: @ben-pr-p  if no message is found `message.send_status` will throw an error won't it?

PR #345, opened 2019-11-03
fix(exports): handle error from stream creation

bchrobot:
An error thrown during creation of the upload stream should be rejected, not left unhandled.


ben-pr-p:
Ooc, what's the effect of this on behavior? To properly clean up upload jobs that fail because of credential issues?

PR #371, opened 2019-11-19
refactor(knex): replace all r.table queries with r.knex

bchrobot:
One important thing to note is that rethink-knex-adapter treats empty strings as `null`s.
See [`_prepSaveFields`][1] for implementation. knex knows to translate `{ column_name: null }` to
`column_name is null`.

I have tested all of the affected queries/mutations except for the two affected `cacheable_queries`
as a) we are not using this functionality, and b) I would first have to spend the time getting up
to speed on Spoke's Redis caching.

The other issue is what to do about `.changes()` (used for notifications). From reading
[`rethink-knex-adapter` documentation][2] and [source code][3], it looks like listeners are
implemented nodejs-side as a pub-sub model arbitrated by `r.table`. So this PR will partially break
user notifications (`ModelName.save({})`) will still trigger changes for now.

[1]: https://github.com/MoveOnOrg/rethink-knex-adapter/blob/master/models.js#L306
[2]: https://github.com/MoveOnOrg/rethink-knex-adapter#changes
[3]: https://github.com/MoveOnOrg/rethink-knex-adapter/blob/master/models.js#L326-L332


src/server/api/lib/fakeservice.js
```diff
@@ -45,22 +46,20 @@ async function convertMessagePartsToMessage(messageParts) {
...
   const { contact_number, user_number, service_id, text } = message;
-  const pendingMessagePart = new PendingMessagePart({
+  const [{ id: partId }] = await r.knex("pending_message_part").insert({
     service: "fakeservice",
     service_id,
     parent_id: null,
     service_message: text,
     user_number,
     contact_number
   });
```
ben-pr-p: Does this need a `.returning('id')`, or is the default `.returning('id')` if nothing else is specified?

src/server/api/lib/nexmo.js
```diff
@@ -89,9 +89,12 @@ async function rentNewCell() {
 }
 
 async function sendMessage(message, trx) {
```
ben-pr-p:
```
  async function sendMessage(message, trx = r.knex) {
```

totally optional because we're in nexmo.js

src/server/api/lib/twilio.js
```diff
@@ -166,6 +166,7 @@ const twilioClient = async messagingServiceSid => {
 };
 
 async function sendMessage(message, organizationId, trx) {
```
ben-pr-p: Prefer `trx = r.knex`

src/server/api/lib/fakeservice.js
bchrobot: Good catch!

bchrobot: Oh, you weren't looking at the most recent commit. This change has already been made

PR #381, opened 2019-12-04
feat: implement unmarkforsecondpass and deleteneedsmessage

ben-pr-p:
Deletion protection:

Screen Shot 2019-12-04 at 1 59 04 PM

### DIFF
--- a/src/containers/CampaignList/OperationDialog.jsx
+++ b/src/containers/CampaignList/OperationDialog.jsx
@@ -21,7 +21,9 @@ export const operations = {
     body: () => `Marking unresponded to messages for this campaign will reset the state of messages that have\
       not been responded to by the contact, causing them to show up as needing a first text, as long as the campaign\
       is not past due. After running this operation, the texts will still be assigned to the same texter, so please\
-      run 'Release Unsent Messages' after if you'd like these second pass messages to be available for auto-assignment.`
+      run 'Release Unsent Messages' after if you'd like these second pass messages to be available for auto-assignment.\
+      \n\
+      Note: you should not run this operation if all initial messages have not yet been sent.`
   },
   releaseUnrepliedMessages: {
     title: campaign => `Release Unreplied Conversations for ${campaign.title}`,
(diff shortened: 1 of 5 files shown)


src/containers/CampaignList/OperationDialog.jsx
```diff
@@ -139,44 +160,81 @@ export const OperationDialogBody = props => {
...
-    clearInProgress,
-    executeOperation
-  } = props;
+const DELETION_PROTECTION_TEXT = "delete contacts";
 
-  const [operationName, campaign] = inProgress;
-  const operationDefinition = operations[operationName];
+export class OperationDialog extends React.Component {
+  state = {
+    pendingDeletionProtectionCheck: false,
```
bchrobot: Not necessary right now as there is only one such operation, but "pending deletion check" would be better as a general "pending operation confirmation check"

src/server/api/schema.js
```diff
@@ -1865,6 +1865,98 @@ const rootMutations = {
...
+      const campaign = await r
+        .knex("campaign")
+        .where({ id: parseInt(campaignId) })
+        .first(["organization_id", "is_archived"]);
+
+      const organizationId = campaign.organization_id;
+
+      await accessRequired(user, organizationId, "ADMIN", true);
+
+      const queryArgs = [parseInt(campaignId)];
```
bchrobot: Why define this on its own? There are no conditional additions to it and it does not seem necessary for readability. Same for `deleteNeedsMessage`

src/server/api/schema.js
```diff
@@ -1865,6 +1865,98 @@ const rootMutations = {
...
+            from message
+            where message.campaign_contact_id = campaign_contact.id
+              and is_from_contact = true
+          )
+        ;
+      `,
+        queryArgs
+      );
+
+      const updateResult = updateResultRaw.rowCount;
```
bchrobot: Destructure above? Same for `deleteNeedsMessage`

src/server/api/schema.js
```diff
@@ -1865,6 +1865,98 @@ const rootMutations = {
...
+        .first(["organization_id", "is_archived"]);
+
+      const organizationId = campaign.organization_id;
+
+      await accessRequired(user, organizationId, "ADMIN", true);
+
+      const queryArgs = [parseInt(campaignId)];
+
+      /**
+       * "UN-Mark Campaign for Second Pass", will only mark contacts as messaged
```
bchrobot: update or remove this comment

bchrobot: err, I meant that this was copy-pasted from `unMarkForSecondPass` into `deleteNeedsMessage`

PR #386, opened 2019-12-10
Fix/order by in message review

ben-pr-p:


ben-pr-p:
You think we should do that even though it's still sometimes updated at (if the contact hasn't been messaged)?

PR #396, opened 2019-12-21
feat: team scoped assignment complete notifications

bchrobot:
This:

- adds missing `preferredTeamId` argument in `schema.js`
- allows knex context to be passed to `allCurrentAssignmentTargets()` to ensure we are looking at current state
- adds team-scoped assignment complete notifications

### DIFF
--- a/src/server/api/assignment.js
+++ b/src/server/api/assignment.js
@@ -542,19 +542,36 @@ export async function myCurrentAssignmentTarget(
   return options ? options[0] : null;
 }
 
-async function notifyIfAllAssigned(type, user, organizationId) {
+async function notifyIfAllAssigned(
+  organizationId,
+  teamsAssignedTo,
+  parentTrx = r.reader
+) {
   if (config.ASSIGNMENT_COMPLETE_NOTIFICATION_URL) {
-    const assignmentTarget = await currentAssignmentTarget(organizationId);
-    if (assignmentTarget == null) {
-      await request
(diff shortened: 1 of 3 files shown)


src/server/api/assignment.js
```diff
@@ -667,13 +692,18 @@ export async function giveUserMoreTexts(
...
+      const { teamId, teamTitle } = team;
+      teamsAssignedTo.set(teamId, teamTitle);
     }
 
     return countUpdated;
   });
 
   // Async function, not awaiting because response to external assignment tool does not depend on it
-  notifyIfAllAssigned(assignmentInfo.type, auth0Id, organizationId);
+  notifyIfAllAssigned(organizationId, teamsAssignedTo, parentTrx).catch(err =>
```
ben-pr-p: I don't think that this should happen within the same transaction

• It's dangerous to extend a transaction without an `await`. In this case, since this function could return before `parentTrx` is used, parentTrx may be called after the transaction has committed. I don't know what would happen under that circumstance!
• If `parentTrx` is rolled back, we shouldn't notify

To ensure that the transaction has committed, we could either:
A) Queue this outside of `giveUserMoreTexts`, passing required info from `giveUserMoreTexts` back to its caller
B) Throw this in a setTimeout with like 15 seconds

B is easier and I think fine – A has the benefit of opting not to run (saving a database round trip, client usage, etc.) if nothing was assigned

bchrobot: I thought about this when I was writing it, but chose the lazy path 😔 

Changes incoming

PR #414, opened 2019-12-24
refactor: update react-router to 5.1.2

bchrobot:
**Motivation**

A piecemeal path to React 16 does exist. This is part of it.

**Description**

The main `react-router` differences are:

- React components are now in react-router-dom package
- router prop has been renamed to history
- params prop has been repalced with match prop
- location.query has been replaced by location.search, which we must parse ourselves
- IndexRoute and IndexRedirect have been replaced with <Route exact={true} />

Additionally:

- This removes the redux Store object as it added complexity and dependencies without adding value
- Authn/Authz have been cleaned up
- TexterDashboard has been cleaned up


src/components/IncomingMessageFilter.jsx
```diff
@@ -367,11 +367,7 @@ IncomingMessageFilter.propTypes = {
   assignmentsFilter: type.shape({
     texterId: type.number
   }).isRequired,
-  tags: type.shape({
-    specificTagIds: type.arrayOf(
-      type.shape({ id: type.string, title: type.string })
-    )
-  }),
+  tags: type.arrayOf(type.shape({ id: type.string, title: type.string })),
```
ben-pr-p: Was this a bug in the prop-types? I don't see anything else changing with respect to the shape here

ben-pr-p:
Can you describe how this is part of a piecemeal migration to React 16?

src/components/IncomingMessageFilter.jsx
bchrobot: Yes, this is fixing the declared PropTypes to be in line with what is actually passed. It is not necessary for `react-router`, but was noise in the console while debugging.

PR #424, opened 2020-01-03
fix: ignore the mocked ID added to CannedResponseInput

bchrobot:


src/server/api/schema.js
```diff
@@ -318,15 +318,13 @@ async function editCampaign(id, campaign, loaders, user, origCampaignRecord) {
...
 
   if (campaign.hasOwnProperty("cannedResponses")) {
-    const cannedResponses = campaign.cannedResponses;
-    const convertedResponses = [];
-    for (let index = 0; index < cannedResponses.length; index++) {
-      const response = cannedResponses[index];
-      convertedResponses.push({
+    // Ignore the mocked `id` automatically created on the input by GraphQL
+    const convertedResponses = campaign.cannedResponses.map(
+      ({ id: cannedResponseId, ...response }) => ({
```
ben-pr-p:
```suggestion
      ({ id: _cannedResponseId, ...response }) => ({
```

Makes it clearer that it's to be ignored, not used

PR #494, opened 2020-01-19
feat: optimize get assignment contacts

ben-pr-p:
This should reduce the number of database round trips from 50+ to 2-4 per request to `getAssignmentContacts`

### DIFF
--- a/src/server/api/campaign-contact.js
+++ b/src/server/api/campaign-contact.js
@@ -161,7 +161,7 @@ export const resolvers = {
         return campaignContact.messages;
       }
 
-      let messages = await r
+      const messages = await r
         .reader("message")
         .where({ campaign_contact_id: campaignContact.id })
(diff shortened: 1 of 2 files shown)


src/server/api/campaign-contact.js
```diff
@@ -160,20 +166,6 @@ export const resolvers = {
         .where({ campaign_contact_id: campaignContact.id })
         .orderBy("created_at");
 
-      // This covers edge case campaign contacts from mid-February 2019
-      if (
-        messages.length === 0 &&
-        campaignContact.message_status !== "needsMessage"
-      ) {
-        messages = await r
```
bchrobot: Change previous declaration from `let messages =` to `const messages =`?

bchrobot: Change previous declaration from `let messages =` to `const messages =`?

PR #497, opened 2020-01-22
Feat/optimize get todos

ben-pr-p:
See commit messages

### DIFF
--- a/src/server/api/assignment.js
+++ b/src/server/api/assignment.js
@@ -9,6 +9,7 @@ import { sleep } from "../../lib/utils";
 import { isNowBetween } from "../../lib/timezones";
 import { r, cacheableData } from "../models";
 import { eventBus, EventType } from "../event-bus";
+import { memoizer, cacheOpts } from "../memoredis";
 
 class AutoassignError extends Error {
   constructor(message, isFatal = false) {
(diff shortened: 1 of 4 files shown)


src/server/api/assignment.js
```diff
@@ -813,16 +814,83 @@ export async function assignLoop(
...
+          }
+        }
+
+        const messageStatusOptions = contactsFilter.messageStatus.split(",");
+        if (!messageStatusOptions.includes(message_status)) {
+          return false;
+        }
+      }
+
+      if ("isOptedOut" in contactsFilter && contactsFilter.isOptedOut != null) {
```
bchrobot: Why not just check for `contactsFilter.isOptedOut` truthiness? Same question for other non-null checks above. Maybe add a comment as the rest of the codebase uses truthiness checks pretty liberally so this is sort of a departure from convention

src/server/api/user.js
```diff
@@ -182,16 +183,63 @@ export const resolvers = {
...
+      const assignments = await r
         .reader("assignment")
-        .select("assignment.*")
-        .join("campaign", "campaign.id", "assignment.campaign_id")
-        .where({
-          "assignment.user_id": user.id,
-          is_started: true,
-          organization_id: parseInt(organizationId),
-          is_archived: false
+        .whereIn("id", todos.rows.map(t => t.assignment_id));
```
bchrobot: pass de-duped set of assignment IDs
```suggestion
        .whereIn("id", Object.keys(shadowCountsByAssignmentId));
```

src/server/api/assignment.js
ben-pr-p: For all of these, they could be either not present or null in which case we don't want to filter on that parameter, or true or false.

PR #498, opened 2020-01-24
fix: tweak todos queries

ben-pr-p:


src/server/api/assignment.js
```diff
@@ -840,7 +840,7 @@ const getContactsCountFromShadowCounts = (shadowCounts, contactsFilter) => {
       if (contactsFilter.messageStatus) {
         if (contactsFilter.messageStatus === "needsMessageOrResponse") {
           if (
-            message_status !== "needsResponse" &&
+            message_status !== "needsResponse" ||
```
bchrobot: doh. I should've caught that the last time around

ben-pr-p: same

PR #500, opened 2020-01-26
feat: cache information for get contacts

ben-pr-p:
Since we don't support `userCannedResponses`, we just return `[]` for them, always.

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -324,11 +324,11 @@ async function editCampaign(id, campaign, loaders, user, origCampaignRecord) {
   }
 
   if (campaign.hasOwnProperty("cannedResponses")) {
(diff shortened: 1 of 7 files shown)


src/server/api/schema.js
```diff
@@ -322,6 +325,10 @@ async function editCampaign(id, campaign, loaders, user, origCampaignRecord) {
 
   if (campaign.hasOwnProperty("cannedResponses")) {
     // Ignore the mocked `id` automatically created on the input by GraphQL
+    memoizer.invalidate(cacheOpts.CampaignCannedResponses.key, {
```
bchrobot: Move this up so the comment about ignoring mocked `id` is still next to where the ignoring is actually happening.

PR #518, opened 2020-02-02
fix: use redis lock to prevent possibility of double assignment

ben-pr-p:


src/server/api/assignment.js
```diff
@@ -590,43 +590,52 @@ export async function fulfillPendingRequestFor(auth0Id) {
...
-    try {
-      const numberAssigned = await giveUserMoreTexts(
-        auth0Id,
-        pendingAssignmentRequest.amount,
-        pendingAssignmentRequest.organization_id,
-        pendingAssignmentRequest.preferred_team_id,
-        trx
-      );
+  const doAssignment = memoizer.memoize(
+    async ({ pendingAssignmentRequestId: _ignore }) => {
```
bchrobot: Why use `pendingAssignmentRequest.id` and not the `pendingAssignmentRequestId ` memoize arg?

src/server/memoredis.js
```diff
@@ -51,7 +51,9 @@ const cacheOpts = {
   OrganizationEscalatedTagList: ["organization-escalated-tag-list", ONE_WEEK],
   OrganizationSingleTon: ["organization-singleton", ONE_HOUR],
   UserOrganizations: ["user-organizations", ONE_WEEK],
-  UserOrganizationRoles: ["user-organization-roles", ONE_WEEK]
+  UserOrganizationRoles: ["user-organization-roles", ONE_WEEK],
+  CampaignOrganizationId: ["campaign-organiation-id", ONE_WEEK],
```
bchrobot: This `CampaignOrganizationId` addition seems to be unrelated. Bad copy-paste or has this been missing and causing issues elsewhere?

ben-pr-p: I think it's from the merge conflict resolution? It should be there

src/server/api/assignment.js
ben-pr-p: No reason! The point of the arg is just to control the locking, etc. It'll get the right pendingAssignmentRequestId from the closer regardless

PR #519, opened 2020-02-03
feat: improve release all unnhandled replies

ben-pr-p:
This PR
– adds an option to releasing all replies that toggles whether to include campaigns that have have limit assignment to teams on or not
– adds loading indicators to releasing all replies, since it could take a bit
– makes ageInHours accept decimal hour interfaces!

### DIFF
--- a/src/containers/AdminCampaignList.jsx
+++ b/src/containers/AdminCampaignList.jsx
@@ -115,6 +115,13 @@ class AdminCampaignList extends React.Component {
         releaseOnRestricted
       )
       .then(result => {
+        if (result.errors) {
+          return this.setState({
+            releaseAllRepliesError: errors,
+            releasingInProgress: false
+          });
+        }
+
         this.setState({
           releaseAllRepliesResult: result.data.releaseAllUnhandledReplies,
           releasingInProgress: false
@@ -242,7 +249,10 @@ class AdminCampaignList extends React.Component {
                 Error: {JSON.stringify(this.state.releaseAllRepliesError)}
               </span>
             ) : this.state.releaseAllRepliesResult ? (
-              <span>Released {this.state.releaseAllRepliesResult} replies</span>
+              <span>
+                Released replies on {this.state.releaseAllRepliesResult}
+                campaigns
+              </span>
             ) : !doneReleasingReplies ? (
               <div>
                 How many hours ago should a conversation have been idle for it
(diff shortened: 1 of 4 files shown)


src/containers/AdminCampaignList.jsx
```diff
@@ -93,20 +94,38 @@ class AdminCampaignList extends React.Component {
...
+    this.setState({ releasingInProgress: true });
 
     this.props.mutations
-      .releaseAllUnhandledReplies(this.props.params.organizationId, ageInHours)
+      .releaseAllUnhandledReplies(
+        this.props.match.params.organizationId,
+        ageInHours,
+        releaseOnRestricted
+      )
       .then(result => {
```
bchrobot: This needs a

```suggestion
       .then(result => {
         if (result.errors) throw result.errors;
         return result
```

src/containers/AdminCampaignList.jsx
```diff
@@ -93,20 +94,38 @@ class AdminCampaignList extends React.Component {
...
         this.setState({
-          releaseAllRepliesResult: result.data.releaseAllUnhandledReplies
+          releaseAllRepliesResult: result.data.releaseAllUnhandledReplies,
+          releasingInProgress: false
```
bchrobot: Do

```
this.setState({ releasingInProgress: false })
```

just once in a final `.then()` to be in the style of try/catch/finally

bchrobot: or make the whole function async and actually use try/catch/finally

src/server/api/question.js
```diff
@@ -36,10 +36,12 @@ export const resolvers = {
     nextInteractionStep: async answer => {
       const getInteractionStep = memoizer.memoize(
         async ({ interactionStepId }) => {
-          return await r
+          const interactionStep = await r
```
bchrobot: accidental `git commit -a`? These should be a separate PR

src/server/api/schema.js
```diff
@@ -2625,50 +2625,60 @@ const rootMutations = {
...
+                select 1 
+                from campaign_contact_tag
+                join tag on tag.id = campaign_contact_tag.tag_id
+                where tag.is_assignable = false
+                  and campaign_contact_tag.campaign_contact_id = campaign_contact.id
+              )
+            returning 1, campaign_id
+          )
+          select
+            count(*) as count, campaign_id
```
bchrobot: drop the `count(*)` as the result only cares about the number of campaigns:

```js
return rawResult.rowCount;
```

src/server/api/question.js
ben-pr-p: Yep

PR #527, opened 2020-02-04
feat: 404 like page for bad campaign id

ben-pr-p:


src/containers/AdminCampaignStats.jsx
```diff
@@ -155,6 +155,11 @@ class AdminCampaignStats extends React.Component {
     const { data, match, adminPerms } = this.props;
     const { organizationId, campaignId } = match.params;
     const campaign = data.campaign;
+
+    if (!campaign) {
+      return <h> Uh oh! This campaign doesn't seem to exist </h>;
```
bchrobot: Needs heading level? Also should include a link back to the campaign list
```suggestion
      return <h1>Uh oh! This campaign doesn't seem to exist.</h1>;
```

PR #531, opened 2020-02-06
feat: release replies on textable contacts only

ben-pr-p:


src/containers/AdminCampaignList.jsx
```diff
@@ -105,19 +105,22 @@ class AdminCampaignList extends React.Component {
...
         ageInHours,
-        releaseOnRestricted
+        releaseOnRestricted,
+        limitToCurrentlyTextableContacts
       )
       .then(result => {
         if (result.errors) {
           return this.setState({
-            releaseAllRepliesError: errors,
+            releaseAllRepliesError: result.errors,
```
bchrobot: I think this would be cleaner in the try/catch/finally pattern

```js
      .then(result => {
        if (result.errors) throw result.errors;

        this.setState({
          releaseAllRepliesResult: result.data.releaseAllUnhandledReplies
        });
      })
      .catch(error => this.setState({ releaseAllRepliesError: error }))
      .then(() =>  this.setState({ releasingInProgress: false }));
```

src/containers/AdminCampaignList.jsx
```diff
@@ -269,6 +272,16 @@ class AdminCampaignList extends React.Component {
                 teams? If unchecked, replies on campaigns restricted to team
                 members will stay assigned to their current texter.
                 <Toggle ref="releaseOnRestricted" defaultValue={false} />
+                <br />
+                <br />
+                Should we limit the releasing to only contacts who are within
```
bchrobot: This is a bit verbose. Strip it down a little to something like

> Release contacts only if it is within texting hours in the contact's timezone? If unchecked, replies will be released for contacts that may not be textable until later today or until tomorrow.

src/containers/AdminCampaignList.jsx
ben-pr-p: I'll do 
```
      .then(result => {
        if (result.errors) {
          throw result.errors;
        }

        this.setState({
          releaseAllRepliesResult: result.data.releaseAllUnhandledReplies,
          releasingInProgress: false
        });
      })
      .catch(error => {
        this.setState({
          releaseAllRepliesError: error,
          releasingInProgress: false
        });
      })
```

No reason to setState and trigger rerenders twice

PR #535, opened 2020-02-06
fix: limit request to current assignment max_request_count

ben-pr-p:


src/server/api/assignment.js
```diff
@@ -737,7 +748,11 @@ export async function assignLoop(
 
   // Determine which campaign to assign to – optimize to pick winners
   let campaignIdToAssignTo = assignmentInfo.campaign.id;
-  let countToAssign = countLeft;
+  let countToAssign = Math.min(
```
bchrobot: It doesn't make sense to me for `assignLoop()` to care about enforcing `max_request_count`. `assignLoop()` is only called from `giveUserMoreTexts()` where the `Math.min()` is already happening. I think removing the export for `assignLoop()` and leaving the `Math.min()` enforcement up to `giveUserMoreTexts()` is the way to go.

ben-pr-p: The difficulty is that assignLoop may be assigning to a different target than giveUserMoreTexts. Without this extra condition, someone could request 500 texts on team A, and then with only 100 left to assign, end up having 400 carry over to team B even though team B's max is 100.

bchrobot: ah, right right

PR #551, opened 2020-02-10
feat: optimize getContactMessagingService

ben-pr-p:
We were joining on the campaign, increasing planning time, to get the organization id, which we already knew!

### DIFF
--- a/src/server/api/lib/message-sending.js
+++ b/src/server/api/lib/message-sending.js
@@ -103,53 +103,50 @@ export const getMessagingServiceById = async messagingServiceId =>
  * @param {number} campaignContactId The ID of the target campaign contact
  * @returns {object} Assigned messaging service Postgres row
  */
-export const getContactMessagingService = async campaignContactId => {
+export const getContactMessagingService = async (
+  campaignContactId,
+  organizationId
+) => {
   if (config.DEFAULT_SERVICE === "fakeservice")
     return { service_type: "fakeservice" };
 
   const {
-    rows: [lookupResult]
+    rows: [existingMessagingService]
   } = await r.reader.raw(
     `
-      with cc_record as (
-        select campaign_contact.cell, campaign.organization_id
-        from campaign_contact
-          join campaign on campaign.id = campaign_contact.campaign_id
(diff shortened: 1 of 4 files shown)


src/server/api/lib/message-sending.js
```diff
@@ -103,53 +103,50 @@ export const getMessagingServiceById = async messagingServiceId =>
...
-        and messaging_service_stick.cell = cc_record.cell
-      ;
+      where exists (
+        select 1
+        from messaging_service_stick
+        where exists (
+            select 1
+            from campaign_contact
+            where messaging_service_stick.cell = campaign_contact.cell
+              and campaign_contact.id = 181403618
```
bchrobot:
```suggestion
              and campaign_contact.id = ?
```

src/server/api/lib/message-sending.js
```diff
@@ -103,53 +103,50 @@ export const getMessagingServiceById = async messagingServiceId =>
...
     throw new Error(`Unknown campaign contact ID ${campaignContactId}`);
 
-  const {
-    cc_organization_id: organization_id,
-    cc_cell: cell,
-    ...existingMessagingService
-  } = lookupResult;
-
-  // Return an existing match if there is one
+  // Return an existing match if there is one - this is the vast majorityu of cases
```
bchrobot:
```suggestion
  // Return an existing match if there is one - this is the vast majority of cases
```

src/server/api/lib/message-sending.js
```diff
@@ -103,53 +103,50 @@ export const getMessagingServiceById = async messagingServiceId =>
...
+          )
+          and organization_id = ?
+      );
     `,
-    [campaignContactId]
+    [campaignContactId, parseInt(organizationId)]
   );
 
-  if (!lookupResult)
+  if (!existingMessagingService)
```
bchrobot: This doesn't make sense anymore. The previous query always returned a record (`right join`) and tested on null fields to see if the messaging service stick existed. The new query should just test on whether a record is returned. I don't think we need to worry about testing if an unknown campaign contact ID was passed?

PR #560, opened 2020-02-13
feat: host-based GraphQL context

bchrobot:
This extends the current GraphQL context to include database and memoizer clients. Currently these will always be the same but could be extended in the future for per-host contexts.

We probably won't do the big schema-based multitenancy refactor but we might as well merge the work from this year-old PR before it goes stale.

Needs to be rebased on `master` after #726 is merged.


henryk1229 (approved):

src/server/contexts.ts
```diff
@@ -0,0 +1,71 @@
...
+import createMemoizer from "memoredis";
+
+import { config } from "../config";
+import logger from "../logger";
+import { createLoaders, r } from "./models";
+import { Memoizer } from "./types";
+
+export interface SpokeDbContext {
+  schema: string;
+  master: Knex;
```
ben-pr-p, line 13:
```suggestion
  primary: Knex;
```
The world seems to be going towards this, we should follow?

ben-pr-p (changes requested):
This is incredibly close to full schema tenancy and looks great overall!

ben-pr-p:
Definitely maybe later!

ben-pr-p (approved):

ben-pr-p:
I think we should do a release after merging this, before merging too much else

PR #592, opened 2020-03-08
fix: restrict numbers to 10 digit US numbers

bchrobot:


ben-pr-p:
We should be using libphonenumber, right? That was previously being used on the client. If that wasn't properly ported over for the server-side refactor, that's should be done instead of this

PR #700, opened 2020-08-01
fix: invalidate cached external lists

bchrobot:
## Description

Clear cached external lists a) when the external system credentials have changed, and b) prior to re-fetching lists.

## Motivation and Context

Many users have been confused by VAN permissions. A common path is to:
1. Create a Spoke API key
1. Share VAN folder with Spoke API key
1. Create Integration with Spoke API key
1. Sync lists from VAN
1. Attempt to load list, but fail due to wrong API key type
1. Create new Hustle API key
1. DO NOT share VAN folder with Hustle API key
1. Re-fetch VAN lists
    - Hustle API key returns 0 lists (no shared folders)
    - Existing external lists from Spoke API key persist
1. Attempt to load list, but fail due to "invalid savedListId in context" -- Hustle API key does not have access to the saved list that the Spoke key fetched

## How Has This Been Tested?

This has been tested locally.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/migrations/20200801094746_clear_external_list_cache_on_refresh.js
+++ b/migrations/20200801094746_clear_external_list_cache_on_refresh.js
@@ -1,0 +1,66 @@
+exports.up = function(knex) {
+  return knex.raw(`
+    -- Delete existing external list records before kicking off fetch
+    create or replace function "public".queue_refresh_saved_lists(van_system_id uuid) returns void as $$
+    declare
+      v_username text;
+      v_api_key_ref text;
+      v_secret graphile_secrets.secrets;
+    begin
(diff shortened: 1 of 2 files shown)


ben-pr-p:
Is it necessary to delete external saved lists in both places (both the inside the job and the mutation)?

I think doing it within the job could cause problems – user A goes to load a saved list, user B refreshes them, and while fetch-saved-list is running, the external list is not in the DB.

Based on my understanding of the issue / solution, just doing it within the mutation should be enough and avoid this race condition?

ben-pr-p:
Ok, sounds good!

PR #722, opened 2020-08-26
feat(van): two-way van sync

bchrobot:
## Description

Support two-way VAN sync.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).


migrations/20200826125123_van_sync_configuration.js
```diff
@@ -0,0 +1,433 @@
...
+
+      return NEW;
+    end;
+    $$ language plpgsql;
+
+    create trigger _500_campaign_external_system_id 
+      before insert or update on campaign 
+      for each row execute procedure tg_campaign_check_exteral_system_id();
+
+    alter table public.all_question_response
```
ben-pr-p, line 32: I'm seeing this already present on production dbs:
```
    "question_response_interaction_step_id_foreign" FOREIGN KEY (interaction_step_id) REFERENCES interaction_step(id)
```

migrations/20200826125123_van_sync_configuration.js
```diff
@@ -0,0 +1,433 @@
...
+      add constraint interaction_step_id_fk foreign key(interaction_step_id) references public.interaction_step(id);
+
+    create unique index all_question_response_value_interaction_step_idx
+      on public.all_question_response(value, interaction_step_id);
```
ben-pr-p, line 36: I have a feeling we're going to see some speed improvements from this one elsewhere as well, and its write frequency is pretty small, so big 👍 !

migrations/20200826075146_van_survey_questions.js
```diff
@@ -0,0 +1,134 @@
...
+    create table public.external_survey_question (
+      system_id uuid not null references external_system(id),
+      external_id integer not null,
+      type text,
+      cycle integer,
+      name text,
+      medium_name text,
+      short_name text,
+      script_question text,
+      status van_data_collection_status,
```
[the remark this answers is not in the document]
bchrobot: I'm on the fence about how specific to make it. The rest of the schema is already pretty VAN-specific but it's still named to support generalization. We briefly talked on the planning call about whether to just commit to VAN and assume we won't be doing much with externals systems versus trying to stick with a more general design but I don't think made a firm decision on that.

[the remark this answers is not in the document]
ben-pr-p: I think it's fine to keep it VAN specific everywhere except enums, pretty much. PostgreSQL enums are hard to change! We could not check in adding a value to an enum in knex migration, for example, since it can't be run in a transaction. No good!

migrations/20200826075146_van_survey_questions.js
```diff
@@ -0,0 +1,134 @@
...
+      foreign key (external_survey_question_id, system_id) references external_survey_question(external_id, system_id)
+    );
+
+    create trigger _500_external_survey_question_response_option_updated_at
+      before update
+      on public.external_survey_question_response_option
+      for each row
+      execute procedure universal_updated_at();
+
+    create function public.insert_van_survey_questions(payload json, result json, context json) returns void as $$
```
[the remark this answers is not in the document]
bchrobot: That's an open ToDo that I added to the bottom of the Notion doc. I think what makes sense is to just delete mappings that are no longer valid. If the campaign has already been synced to VAN then VAN will have had to deal with existing recorded responses. If the campaign has not been synced yet (or will be synced again) then deleting the **_may_** force the user to re-configure mappings. Maybe it makes sense to delete entire config object though if one of its mappings becomes invalid to force re-configuring and prevent accidentally syncing partial data?

sql```
-- DDL for tables referencing external_survey_question:

  foreign key (external_survey_question_id, system_id) references external_survey_question(external_id, system_id) on delete cascade;

-- update for public.insert_van_survey_questions()
with insert_results as (
    -- ...
    returning id
)
delete from external_survey_question
where
  system_id = ?,
  external_id = ?
  and id not in insert_results
;
```

File: migrations/20200826122046_van_result_codes.js [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+      on conflict (system_id, external_id)
+      do update set
+        name = EXCLUDED.name,
+        medium_name = EXCLUDED.medium_name,
+        short_name = EXCLUDED.short_name
+      ;
+    end;
+    $$ language plpgsql volatile security definer set search_path = "public";
+
+    create function public.queue_refresh_van_result_codes(van_system_id uuid) returns void as $$
```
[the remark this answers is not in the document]
bchrobot: If result codes cannot be archived then we should be clearing the table completely before inserting new ones? That would break existing mappings

migrations/20200826125123_van_sync_configuration.js
```diff
@@ -0,0 +1,433 @@
+exports.up = function(knex) {
+  return knex.raw(`
+    alter table public.campaign
+      add column external_system_id uuid references public.external_system(id);
+
+    create index campaign_external_system_id_index on campaign(external_system_id);
```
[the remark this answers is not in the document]
bchrobot: true, `external_system` is small enough that it'll probably always be a sequential scan, even for instances with _many_ organizations

migrations/20200826125123_van_sync_configuration.js
bchrobot: I wonder how my dev DB got out of sync 🤔

ben-pr-p:
> > We should add a write to `campaign.external_system_id` on list import, and that should be the only place from which the user can write to it
> 
> It feels a little weird to jam that into the contact selection. The current UI changes prevent VAN list loading unless `campaign.external_system_id` is set from the standalone campaign builder section. Maybe that achieves the same goal? In which case we just need to add a step to CSV list loading that sets `campaign.external_system_id = null`

Yeah, I think that's right: current behavior + external_system_id = null on CSV list load

migrations/20200826125123_van_sync_configuration.js
bchrobot: For reference, this index is created here:
https://github.com/politics-rewired/Spoke/blob/master/migrations/20190207220000_init_db.js#L206-L207

bchrobot: I didn't register the table -> view switch and thought it was getting dropped somewhere else. Took a quick look for where that was happening resulting in:
https://github.com/politics-rewired/Spoke/pull/728

migrations/20200826075146_van_survey_questions.js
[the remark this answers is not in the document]
bchrobot: Addressed here:
https://github.com/politics-rewired/Spoke/pull/722/commits/919dd59d42b38d9b54d422f125ca06e8228d0949#diff-106549701d17cd48abf3389d06dec13bR47-R116

migrations/20200826122046_van_result_codes.js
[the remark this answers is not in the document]
bchrobot: Addressed here https://github.com/politics-rewired/Spoke/pull/722/commits/919dd59d42b38d9b54d422f125ca06e8228d0949#diff-106549701d17cd48abf3389d06dec13bR47-R116 (and with `on delete cascade` on fks)

migrations/20200826111414_van_activist_codes.js
```diff
@@ -0,0 +1,85 @@
...
+      primary key (external_id, system_id)
+    );
+
+    create trigger _500_external_activist_code_updated_at
+      before update
+      on public.external_activist_code
+      for each row
+      execute procedure universal_updated_at();
+
+    create function public.insert_van_activist_codes(payload json, result json, context json) returns void as $$
```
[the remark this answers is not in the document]
bchrobot: Here: https://github.com/politics-rewired/Spoke/pull/722/commits/919dd59d42b38d9b54d422f125ca06e8228d0949#diff-106549701d17cd48abf3389d06dec13bR47-R116

src/server/tasks/sync-campaign-contact-to-van.ts
```diff
@@ -0,0 +1,223 @@
...
+      `,
+      [contactId]
+    )
+    .then(({ rows }: { rows: CanvassResultRow[] }) => rows);
+
+  if (canvasResultsRaw.length === 0) return;
+
+  let chain = Promise.resolve();
+  for (const canvassResult of canvasResultsRaw) {
+    chain = chain.then(async () => {
```
ben-pr-p, line 176: I think just putting the `await` inside the for loop is easier to read that the chain pattern.

Shouldn't we just be making one request per contact though?

migrations/20200915113240_van-sync-task.js
```diff
@@ -4,28 +4,43 @@ exports.up = function(knex) {
...
+      select
+        (context->>'job_request_id')::integer,
+        (context->>'contact_count')::integer
+      into v_job_request_id, v_contact_count;
+
+      with locked_rows as (
+        select 1
+        from graphile_worker.jobs jobs
+        where (jobs.payload->'__context'->>'job_request_id')::integer = v_job_request_id
+        for update
```
ben-pr-p, line 21: This will block the jobs from executing for the duration of this function - I don't think we want that (I think it's better to have slightly inaccurate system progress), since this could be slow because of a delivery report backlog, etc.

migrations/20200915113207_van_survey_questions.js
```diff
@@ -138,7 +138,8 @@ exports.up = function(knex) {
           'api_key', json_build_object('__secret', v_api_key_ref),
           'van_system_id', queue_refresh_van_survey_questions.van_system_id,
           '__after', 'insert_van_survey_questions'
-        )
+        ),
+        'van-api'
```
ben-pr-p, line 142: This could cause a refresh to be delayed while a 100k sync runs - let's have:
`van-api-refresh` and `van-api-sync`

migrations/20200915113210_van_activist_codes.js
```diff
@@ -80,7 +80,8 @@ exports.up = function(knex) {
           'api_key', json_build_object('__secret', v_api_key_ref),
           'van_system_id', queue_refresh_van_activist_codes.van_system_id,
           '__after', 'insert_van_activist_codes'
-        )
+        ),
+        'van-api'
```
ben-pr-p, line 84: This could cause a refresh to be delayed while a 100k sync runs - let's have:
`van-api-refresh` and `van-api-sync`

migrations/20200915113220_van_result_codes.js
```diff
@@ -68,7 +68,8 @@ exports.up = function(knex) {
           'api_key', json_build_object('__secret', v_api_key_ref),
           'van_system_id', queue_refresh_van_result_codes.van_system_id,
           '__after', 'insert_van_result_codes'
-        )
+        ),
+        'van-api'
```
ben-pr-p, line 72: This could cause a refresh to be delayed while a 100k sync runs - let's have:
`van-api-refresh` and `van-api-sync`

migrations/20200915113240_van-sync-task.js
```diff
@@ -111,7 +111,8 @@ exports.up = function(knex) {
               'job_request_id', v_job_request_id,
               'contact_count', v_contact_count
             )
-          )
+          ),
+          'van-api'
```
ben-pr-p, line 115: This could cause a refresh to be delayed while a 100k sync runs - let's have:
`van-api-refresh` and `van-api-sync`

src/containers/AdminExternalSystems.tsx
```diff
@@ -180,7 +180,7 @@ class AdminExternalSystems extends Component<Props, State> {
                     onClick={this.makeStartEditExternalSystem(system.id)}
                   />
                   <RaisedButton
-                    label="Pull lists"
+                    label="Get Resources"
                     labelPosition="before"
                     icon={<SyncIcon />}
```
ben-pr-p, line 185: This SyncIcon looks like a Download icon!

src/containers/AdminExternalSystems.tsx
```diff
@@ -156,7 +156,7 @@ class AdminExternalSystems extends Component<Props, State> {
             <TableRow>
               <TableHeaderColumn>Name</TableHeaderColumn>
               <TableHeaderColumn>Type</TableHeaderColumn>
-              <TableHeaderColumn>Last Synced</TableHeaderColumn>
+              <TableHeaderColumn>Resources Last Fetched</TableHeaderColumn>
```
ben-pr-p, line 159:
```suggestion
              <TableHeaderColumn>Sync Options Last Fetched</TableHeaderColumn>
```

src/containers/AdminExternalSystems.tsx
```diff
@@ -180,7 +180,7 @@ class AdminExternalSystems extends Component<Props, State> {
                     onClick={this.makeStartEditExternalSystem(system.id)}
                   />
                   <RaisedButton
-                    label="Pull lists"
+                    label="Get Resources"
```
ben-pr-p, line 183:
```suggestion
                    label="Refresh Sync Options"
```

ben-pr-p (changes requested):

src/server/tasks/sync-campaign-contact-to-van.ts
bchrobot: I think I did the chaining because of eslint rules. #726 will address this. eslint/prettier is pretty much borked for me locally at the moment

We make one request per `canvassedDate`. I confirmed with Max at EveryAction that `question_response.created_at` is preferred to `campaign_contact.created_at`.

migrations/20200915113240_van-sync-task.js
bchrobot: Without the locking this will often never actually finish because the last 1-3 jobs all think there are others after them.

We could move the locking to the `job_request` record and manipulate counts in the `payload` there but I don't think that really helps. Some kind of locking is required to ensure `__after`s run sequentially.

migrations/20200915113207_van_survey_questions.js
bchrobot: Yeah, it sounds like that's the rub given the throttling they're imposing:

> No more than 2 calls per second per API key for /findOrCreate and **no more than 5 calls per second for other endpoints.**
> No more than 3 concurrent requests per API key/client, throttled by duration.
> -- https://everyaction.readme.io/docs/throttling-guidelines-2020-election

src/containers/AdminExternalSystems.tsx
bchrobot: Is there an alternate icon you would like to suggest? I've already spent a while looking through the available ones

ben-pr-p: How about `Cached`? I'm searching here: https://material-ui.com/components/material-icons/

migrations/20200915113207_van_survey_questions.js
ben-pr-p: Oh ok, then we should use the graphile_worker priority feature to make resource refresh requests go before sync requests. Lower priority goes first

src/containers/AdminExternalSystems.tsx
bchrobot: That was previously rejected because people thought it gave the impression of syncing data back to VAN which I think is true.

migrations/20200915113207_van_survey_questions.js
bchrobot: Resolved in b4354b5

migrations/20200915113210_van_activist_codes.js
bchrobot: Resolved in b4354b5

migrations/20200915113220_van_result_codes.js
bchrobot: Resolved in b4354b5

migrations/20200915113240_van-sync-task.js
bchrobot: Resolved in b4354b5

src/server/tasks/sync-campaign-contact-to-van.ts
bchrobot: Addressed in 1bda114

migrations/20200915113240_van-sync-task.js
bchrobot: Refactored as cronjob in dfb1923

PR #723, opened 2020-08-27
fix: export for van pagination error

ben-pr-p:
Fixes a bug in the export for VAN pagination

## Description

Because we were using campaign_contact.id based pagination, but the outer query was not ordered by campaign_contact.id, we were iterating over some batches of contacts multiple times.

The removal of the distinct on was necessary to overcome [the issue described here](https://stackoverflow.com/questions/9795660/postgresql-distinct-on-with-different-order-by) - there are other solutions if we want it back, but my understanding of the VAN import process is that if there are multiple survey responses for a contact, we actually want multiple rows for that contact.

## Types of changes

## Checklist:


ben-pr-p:
I think if the goal is to have one row per response, we want to join on the the union of question_responses and tags, right?

We also should not join on all_question_response - that includes deleted things

PR #725, opened 2020-09-03
fix(assignable_campaigns): add due_by clause

ben-pr-p:
This change prevents assignment to overdue campaigns. We had already intended this to be the correct behavior, but it appears had not implemented it.

## How Has This Been Tested?

I've tested the view definition locally as well as the correctness of the specific date_trunc / interval semantics on several production databases.

## Types of changes

## Checklist:

- [ ] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/migrations/20200903111251_exclude-overdue-campaigns-from-autoassignment.js
+++ b/migrations/20200903111251_exclude-overdue-campaigns-from-autoassignment.js
@@ -1,0 +1,27 @@
+
+exports.up = function(knex) {
+  return knex.schema.raw(`
+    create or replace view assignable_campaigns as (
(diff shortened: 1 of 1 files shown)


bchrobot (changes requested):
It seems like this should instead be a change to the `assignable_campaigns` view to also filter on `due_by`?

ben-pr-p:
> It seems like this should instead be a change to the `assignable_campaigns` view to also filter on `due_by`?

Oops, forgot to update the PR description. We want replies to be assignable after due date but not initials, hence the second commit overriding the first.

bchrobot (approved):

PR #733, opened 2020-09-23
Fix: Add Tag to Contacts on Save

henryk1229:
## Description

This PR changes the way tags are added to contact conversations, refactoring the handleApplyTags method in the AssignmentTexterContact component so that it runs the tagContact mutation when the user clicks the 'save' button, as opposed to running it as an async function when the user clicks the 'send' button

## Motivation and Context

Currently, tags are added to conversations when a user clicks send message within the AssignmentTexterContact component. If a user applies a tag and navigates away from the component without sending a message, the tag isn't applied.

## How Has This Been Tested?

Behaviorally. 

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/components/AssignmentTexter.jsx
+++ b/src/components/AssignmentTexter.jsx
@@ -16,6 +16,7 @@ import { loadData } from "../containers/hoc/with-operations";
 import AssignmentTexterContact from "../containers/AssignmentTexterContact";
 import Empty from "../components/Empty";
 import LoadingIndicator from "../components/LoadingIndicator";
+import { catchError } from "../network/utils";
 
 const SEND_DELAY = 100;
 
@@ -278,13 +279,6 @@ class AssignmentTexter extends React.Component {
 
     const promises = [];
 
-    const catchError = response => {
-      if (response.errors) {
-        throw new Error(response.errors);
-      }
-      return response;
-    };
-
     if (payload.message)
       promises.push(
         this.props.mutations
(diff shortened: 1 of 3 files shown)


src/components/AssignmentTexter.jsx
```diff
@@ -336,6 +328,16 @@ class AssignmentTexter extends React.Component {
     }
   };
 
+  addTagToContact = (contact_id, tag) => {
+    const catchError = response => {
+      if (response.errors) {
+        throw new Error(response.errors);
+      }
+      return response;
+    };
```
bchrobot, line 337: This is used a bunch of places on the frontend. Move this to a new `src/network/utils.ts` so it can be reused elsewhere?

src/components/AssignmentTexter.jsx
```diff
@@ -522,23 +527,26 @@ const mutations = {
...
-    mutation: gql`
-      mutation tagConversation(
-        $campaignContactId: String!
-        $tag: ContactTagActionInput!
-      ) {
-        tagConversation(campaignContactId: $campaignContactId, tag: $tag) {
-          id
-          assignmentId
+  tagContact: ownProps => (campaignContactId, tag) => {
+    console.log("tag contact runs", campaignContactId, tag, ownProps);
```
bchrobot, line 531: Remove the log and revert to returning an object literal

bchrobot (changes requested):
Welcome to Spoke! Woohooo! 🙌

I'm wondering whether the `{ addedTags, removedTags, pendingNewTags }` state can be dropped from `src/containers/AssignmentTexterContact/index.jsx` entirely and replaced with an Apollo cache update after the mutation.

That may be more trouble than it's worth if the TagPicker modal is a controlled component and `AssignmentTexterContact` does need to hold the state.

bchrobot (approved):

PR #734, opened 2020-09-23
fix(assignment): dont delete assignments, breaks opt out foreign keys

ben-pr-p:
Delete assignment breaks the opt out foreign keys! We can just leave them hanging around at 0.

## Types of changes

## Checklist:


bchrobot (approved):
Ah, so the problem is using this to manage texters after the campaign has begun!

PR #738, opened 2020-10-02
feat(exports): move export to graphile worker

henryk1229:
## Description
This pr changes the way we handle exporting campaigns. ExportCampaign and exportForVan are now 'tasks' managed via graphile worker, as opposed to being handled by the previously defined workers.

## Motivation and Context
As requested by Ben Packer

## How Has This Been Tested?
Behaviorally

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:


bchrobot (changes requested):
But what will I do with all the extra hours in my day where I'm not clearing stuck campaign exports for people?

There's a lot going on in pre-graphile Spoke job world. The `job_request` table is used as the queue and for recording client-facing status and result messages. We want graphile as the queue but we also want to keep the client-facing status and result messages (as graphile jobs are deleted upon completion).

What I think we ultimately want is a task wrapper and an `addJob()` wrapper that manage a `job_request` record. I've moved towards that with the VAN PRs but haven't formalized it (VAN sync is tough because it's a fan-out/fan-in collection of tasks).

This looks something like
```ts
export const addProgressJob = async (identifier: string, payload: any, initialResult: any = {}, context: any = {}) => {
  const [jobResult] = await r
    .knex("job_request")
    .insert({
      job_type: identifier,
      status: 0,
      payload: JSON.stringify(context),
      result_message: JSON.stringify(initialResult)
    })
    .returning("*");

  const wrappedPayload = { ...payload, _jobRequestId: id };
  await getWorker().addJob(identifier, wrappedPayload);
  return jobResult;
};

interface ProgressTaskPayload {
  _jobRequestId: string;
}

interface ProgressTaskOptions {
  removeOnComplete: boolean;
}

// maybe with some better use of generics for __jobRequestId
export const wrappedProgressTask: Task = (
  task: Task,
  options: ProgressTaskOptions
) => (payload, helpers) => {
  const jobId = payload.__jobRequestId;
  const updateStatus = async (status: number) => await r.knex("job_request").update({ status }).where({ id: jobId });
  const updateResult = async (result: any) => await r.knex("job_request").update({ status }).where({ result_message: result });
  const progressHelpers = { ...helpers, updateStatus, updateResult };

  try {
    await task(payload, progressHelpers);

    if (options.removeOnComplete) {
      await r.knex("job_request").where({ id: jobId }).del();
    }
  } catch (err) {
    const { attempts, maxAttempts } = helpers.job;
    if (attempts === maxAttempts) {
      // "Complete" graphile job by recording error on the final attempt
      return updateResult({ error: err.message });
    }
    throw err;
  }
};
```

Used like:

```ts
// worker.ts
m.taskList!["export-campaign"] = wrappedProgressTask(exportCampaign, { removeOnComplete: true });

// export-campaign.ts
await addProgressJob("export-campaign", payload, );
```

Does that make sense? Happy to chat more about

ben-pr-p:
Maybe we want to finish this and add a generic task wrapper as a separate PR?

ben-pr-p:
Also, we should formalize that exports are only run 1 at a time via a `queue_name`. I think running multiple at once is the source of most errors.

bchrobot (approved):

PR #746, opened 2020-10-14
feat: populate tags in dropdown

henryk1229:
## Description
This pr uses react-select to populate conversation tags via dropdown menu within the texter modal. It maintains tag searchability, in the case that the list of applicable tags is too long to navigate and select via the dropdown.

## Motivation and Context
Per Kohlee's request, this will make applying tags to conversations more intuitive, reducing the need for training users, and won't require that texters know the name of a specific tag to search for and apply it.

## How Has This Been Tested?

Behaviorally 
## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:


henryk1229:
> We're using `react-select` successfully in a few other places ([`SelectExcludeCampaigns`](https://github.com/politics-rewired/Spoke/blob/18c0853c1aeac2b906d727cf49cd2dfbea9f3ebd/src/containers/AdminCampaignEdit/sections/CampaignContactsForm/components/SelectExcludeCampaigns.tsx#L52-L61), [`IncomingMessageActions`](https://github.com/politics-rewired/Spoke/blob/b12a35f3f8a31f07aedd345decf09aa32ecd23e5/src/components/IncomingMessageActions.jsx#L135-L142)) so React v15 doesn't seem like a total dealbreaker. Is the issue the z-index you showed last Friday?
> 
> It seems like the [`components`](https://react-select.com/components#replacing-components) prop may be able to solve the problem? [This GH comment](https://github.com/JedWatson/react-select/issues/2208#issuecomment-368374946) shows using `components` to take advantage of portals (from before portals were officially introduced). Then BYO-portal with [`react-portal`](https://www.npmjs.com/package/react-portal) (which has React v15 fallback support) as hinted at in this [SO post](https://stackoverflow.com/a/55468574)?

The issue as I understood it was that react-select was using portals under the hood to render the dropdown menu component when used as part of a modal - I tried to work around it for a while by adjusting the zIndex in various places, to no avail. 

I played around with a custom components prop, but didn't think about a byo portals solution. That GH comment looks like it will work for our needs - I'll refactor for tomorrow.

henryk1229:
i think this should do it!

src/components/TagSelector/MenuPortal.jsx
```diff
@@ -0,0 +1,107 @@
...
+  return r;
+};
+
+export default class MenuPortal extends PureComponent {
+  static propTypes = {
+    menuPlacement: PropTypes.oneOf(["auto", "bottom", "top"]),
+    appendTo: PropTypes.instanceOf(Element),
+    controlElement: PropTypes.instanceOf(Element),
+    menuPosition: PropTypes.oneOf(["absolute", "fixed"])
+  };
```
bchrobot, line 52: oooh

src/components/TagSelector/MenuPortal.jsx
```diff
@@ -0,0 +1,107 @@
...
+    const { offsetLeft, offsetTop } = appendTo;
+    containerLeft = left - offsetLeft;
+    containerTop = top - offsetTop;
+  }
+
+  // get coordinates for dialog modal
+  const rect = controlElement.getBoundingClientRect();
+
+  // calculate position for dropdownMenu
+  const r = {
```
bchrobot, line 28: prefer descriptive variable names

src/components/TagSelector/MenuPortal.jsx
```diff
@@ -0,0 +1,107 @@
...
+    // same wrapper element whether fixed or portalled
+    const menuWrapper = (
+      <div className="menuPortalWrapper">
+        <div style={menuPortalCSS(state)}>{children}</div>
+      </div>
+    );
+
+    // override createPortal
+    const createPortal = child => {
+      const node = document.getElementById("portalRoot");
```
bchrobot, line 100: Where does `portalRoot` come from?

I see the addition of a `<div id="applyTagDialog">` in `ApplyTagDialog.jsx` but it looks like `applyTagDialog` is also unreferenced.

src/components/TagSelector/index.jsx
```diff
@@ -55,57 +56,49 @@ class TagSelector extends Component {
...
+    tagsArray,
+    { action: selectAction, option: selectedTag }
+  ) => {
+    if (selectAction === "select-option") {
+      this.handleSelectTag(selectedTag);
+    }
+    if (selectAction === "remove-value") {
+      this.handleRemoveTag(tagsArray);
+    }
+  };
```
bchrobot, line 70: I think this can be simplified to just call `this.props.onChange()` with some mapped version of the `tagsArray` argument.

Then you can get rid of:

- `handleBeforeRequestAdd()`
- `handleRequestAddTag()`
- `handleRemoveTag ()`
- `addTag()`
- `handleSelectTag()`
- `this.state. pendingTag`
- `<ApplyTagConfirmationDialog />` -- looking over the existing code I'm unclear under what circumstances this was even supposed to be displayed..
- `handleCancelConfirmTag()`
- `handleConfirmAddTag ()`

bchrobot (changes requested):
I played around with it locally and of man does it feel much nicer!

Also, commit the changes to `yarn.lock`

src/components/TagSelector/MenuPortal.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+  if (appendTo) {
+    const { left, top } = appendTo.getBoundingClientRect();
+    const { offsetLeft, offsetTop } = appendTo;
+    containerLeft = left - offsetLeft;
+    containerTop = top - offsetTop;
+  }
+
+  console.log("containerTop".containerTop);
+
```
bchrobot, line 25:
```suggestion
```

bchrobot (approved):

PR #748, opened 2020-10-19
refactor: use unique constraint to prevent duplicates

bchrobot:
## Description

Replace potentially-problematic de-duplication query with Postgres constraint.

## Motivation and Context

We do not allow texters to add canned responses from the UI. The GraphQL mutation could still be called, however. The previous delete duplicates query shouldn't have caused mass-deletion of canned responses as it was user-scoped but better to prevent duplicates using constraints.

## How Has This Been Tested?

This mutation is not currently used in our fork and thus was not tested.

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -1490,32 +1490,21 @@ const rootMutations = {
       return { id };
     },
 
-    createCannedResponse: async (_, { cannedResponse }, { user, loaders }) => {
-      authRequired(user);
+    createCannedResponse: async (_, { cannedResponse }, { user }) => {
+      const campaignId = parseInt(cannedResponse.campaignId, 10);
+      const { organization_id } = await r
+        .knex("campaign")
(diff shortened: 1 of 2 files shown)


src/server/api/schema.js
```diff
@@ -1490,32 +1490,21 @@ const rootMutations = {
...
+        .first(["organization_id"]);
+      await accessRequired(user, organization_id, "TEXTER");
 
       await r.knex("canned_response").insert({
-        campaign_id: cannedResponse.campaignId,
+        campaign_id: campaignId,
         user_id: cannedResponse.userId,
         title: cannedResponse.title,
         text: cannedResponse.text
       });
```
ben-pr-p, line 1506: Do we need an on conflict do nothing here?

bchrobot: I thought about that but I think we do want it to throw an error

PR #790, opened 2020-11-25
fix(escalationtags): check tags in assignment row chip input component

henryk1229:
## Description

This fixes an issue raised by this [ticket](https://secure.helpscout.net/conversation/1343368005/3413?folderId=4118236), where the Chip Input component allows creation of new tags via text input, which aren't currently being checked against escalationTagList. When the user tries to save the new tag, it generates an error - even if the user is entering a tag via text that should be selectable.

## Motivation and Context

Chip Input default behavior is leading to confusion among users about how they can add tags within the AssignmentRow component, as well as what kind of tags they can add. This resolves that confusion in a minimal way.

## How Has This Been Tested?

Behaviorally 

## Types of changes

## Checklist:

### DIFF
--- a/src/containers/AdminAssignmentControl/AssignmentRow.jsx
+++ b/src/containers/AdminAssignmentControl/AssignmentRow.jsx
@@ -42,6 +42,20 @@ const AssignmentRow = props => {
     onChange({ escalationTags: newEscalationTags });
   };
 
+  const handleCheckEscalationTag = newTag => {
+    // when a tag is added by search, newTag will be a string,
+    // not the tag object within escalationTagList -
+    // here we trim possible whitespace, then find and verify the tag
+    // before calling handleAddEscalationTag
+    const trimmedTag = newTag.id ? newTag : newTag.trim();
+    const foundTag = escalationTagList.find(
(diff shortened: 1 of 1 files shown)


src/containers/AdminAssignmentControl/AssignmentRow.jsx
```diff
@@ -42,6 +42,20 @@ const AssignmentRow = props => {
...
+    // here we trim possible whitespace, then find and verify the tag
+    // before calling handleAddEscalationTag
+    const trimmedTag = newTag.id ? newTag : newTag.trim();
+    const foundTag = escalationTagList.find(
+      tag => tag.title === trimmedTag || tag.title === trimmedTag.title
+    );
+    if (foundTag) {
+      handleAddEscalationTag(foundTag);
+    }
+  };
```
bchrobot, line 57: It looks like this should [return a boolean](https://github.com/TeamWertarbyte/material-ui-chip-input/tree/legacy): `return foundTag !== undefined`

bchrobot (changes requested):

bchrobot (approved):

PR #807, opened 2020-12-15
refactor: remove moment.js and use luxon instead

lh00000000:
all usage of moment.js removed in favor of luxon
some types added as well 

## Description
e.g. 
```js
import moment from "moment-timezone";
moment(xyz).fromNow()
```
>>
```ts
import { DateTime } from "luxon";
DateTime.fromJSDate(xyz).toRelative()
```

## Motivation and Context
Closes #755

## How Has This Been Tested?
mostly through locally running ui

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Documentation Changes

no

## Checklist:

- [x via lint-staged ] My code follows the code style of this project.
- [x...or it will ] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] My change requires a change to the documentation.
- [ ] I have included updates for the documentation accordingly.

PR #809, opened 2020-12-16
improvement: sort campaign interaction steps

### LINKED ISSUE #488: getStartingMessageText() for initial messages shows question response script
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**

The only time the message field should start out pre-populated is when sending initials during which it should use the script from the top most script stage. Sometimes it uses the script from a non-root interaction step.

**To Reproduce**
Cannot reliably reproduce.

**Expected behavior**

The root interaction step's script should always be used for initial messages.

**Screenshots**

image-2

**Desktop (please complete the following information):**

The two users where this happened were both using the latest version of Chrome

**Additional context**

Stopgap solution: #485

### LINKED ISSUE #496: Inconsistent interaction step order on "edit campaign" page
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**

The order of interaction steps sharing a parent on the campaign edit page is non-deterministic.

**To Reproduce**
1. Create campaign
2. Add root interaction step
3. Add two child interaction steps
4. Save interaction steps
5. Refresh page a few times
6. See non-deterministic ordering

**Expected behavior**

Interaction steps with the same parent step should be ordered by id (order of creation).

**Additional context**

N/A

lh00000000:
make interaction steps ordered deterministically

## Description
add sorting to children interaction steps

## Motivation and Context
Closes #496 and closes #488

## How Has This Been Tested?
1. create campaign
2. create some steps
3. save the campaign
4. edit the campaign
5. add a step
6. save the campaign
7. note the order 
8. leave the campaign editing dialog
9. return and edit the campaign
10. note the interaction step order

## Types of changes

## Documentation Changes

no

## Checklist:


lh00000000:
> This should be rebased on `master` a) to address file conflicts and b) to exclude commits from #807.
> 
> Squash merges can confuse `git rebase` so the easiest thing to do is often to get surgical with [`--onto`](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#_more_interesting_rebases):
> 
> ```shell
> git checkout master && git pull
> git checkout improvement/sort-interaction-steps
> # bf30da4 is the last commit of the luxon PR
> git rebase -i bf30da4 --onto master
> ```
> 
> Which translates to, "take the `improvement/sort-interaction-steps` branch, figure out the patches since it diverged from the `bf30da4` commit, and replay these patches in the `improvement/sort-interaction-steps` branch as if it was based directly off the `master` branch instead."
> 
> Description should be updated to better link the GH issue:
> 
> ```diff
>   ## Motivation and Context
> - #496
> + Closes #496
> ```

Ah sorry, I forgot I was planning to change the PR on github to say the PR was into the luxon PR, (so the commits will only differ from there i.e. planning to merge that first). I haven't used `--onto` before though so thanks for the guide (maybe I'll end up doing that anyways the luxon PR won't be merged soon (which I'm trying to avoid bc I think rebasing that on to master w another week of changes on master will not turn out well for me))

src/lib/interaction-step-helpers.js
```diff
@@ -89,17 +92,27 @@ export function getTopMostParent(interactionSteps, _isModel) {
...
-  const children = interactionSteps.filter(
-    (is) => is.parentInteractionId === root.id
-  );
+export function makeTree(interactionSteps, id = null, indexed = null) {
+  const indexedById =
+    indexed ||
+    _(interactionSteps)
+      .map((is) => [is.id, is])
+      .fromPairs()
+      .value();
```
bchrobot, line 101: Prefer importing only the lodash util actually needed to keep bundle size down (webpack's tree shaking isn't able to do this automagically):

```ts
import fromPairs from "lodash/fromPairs";
 
const indexedById = indexed || fromPairs(interactionSteps.map((is) => [is.id, is]));
```

Same below

bchrobot (changes requested):
`createdAt` is used over `id` because of the random ID generation client-side? If so include a comment near the `sortBy`

src/lib/interaction-step-helpers.js
lh00000000: what about `const { chain: _ } = require("lodash");`? i've found that picking out pieces from lodash usually leads to using the standard js api and generally more mutable temp-variables code, which takes longer to read and generally adds up more than the hit in sending the bundle over network (which is an intensely cached world anyways)

lh00000000:
> `createdAt` is used over `id` because of the random ID generation client-side? If so include a comment near the `sortBy`

i think i changed this to 
```js
      _(interactionSteps)
      .filter((is) => is.parentInteractionId === root.id)
      .sortBy((is) => {
        const asDate = DateTime.fromISO(is.createdAt);
        return asDate.isValid ? asDate : null;
      })
      .reverse() // ui puts newest step at top
      .map((c) => makeTree(interactionSteps, c.id, indexedById))
      .value()
  };
  ```
createdAt should be generated clientside. i was falling back to id before, but i think i was assuming id were uuids before. i still don't think they should be used as the sort key even though they're backed by an postgres ascending sequence though because they initially have a throwaway in the client then require a db trip get "the real id", so i still think it should be `createdAt` if possible, and then basically giving up in `createdAt` doesn't exist (because theoretically it should always exist)

src/lib/interaction-step-helpers.js
ben-pr-p: I prefer the chain style programming in general as well but don't think it's worth importing all of lodash for it, especially for just this one code block

lh00000000: agreed lol. there were a bunch of things i more-heavily refactored while working on the moment->luxon PR that i took out of the branch later on that were mostly slimming down `let a = []; Object.something(b).forEach(x => a.push(b)` stuff so the suggestion was less "lodash would be good for situation"  and more "maybe we should take the 3mb plunge now, while the sprint is still on and we're all touching a lot of files`. 
alt suggestion is `lodash/fp` this https://medium.com/bootstart/why-using-chain-is-a-mistake-9bc1f80d51ba

lh00000000:
@bchrobot can you take another 2sec glance? i want to be sure i didn't miss anything merging with your `CampaignInteractionStepsForm` rewrite

bchrobot (approved):

PR #813, opened 2020-12-17
feat(van): add operation mode select to integration config

### LINKED ISSUE #787: Add VoterFile/MyCampaign select to VAN integration config

To use the VAN integration in MyCampaign mode you need to manually specify the `|1` API key suffix. This is not advertised and any authorization errors we encounter are not exposed to the user.

There should be a dropdown select widget for selecting Voterfile vs MyCampaign operation mode. Authorization errors for failed syncs should be displayed somewhere on the integration page rather than failing silently.

Add text instructions for having the user add the `|1` suffix. This is poor UX.

**Additional context**

N/A

henryk1229:
## Description

Closes #787, part of Q4 feature freeze.

## Motivation and Context

This allows users to choose between VoterFile and MyCampaign mode when integrating with VAN, something they hadn't been able to explicitly.

It also displays auth errors on failing syncs, instead of swallowing them.

## How Has This Been Tested?

This has been tested locally.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:


bchrobot (commented):
The `operation_mode` column isn't needed with that operation mode already being stored directly in the key.

You can use some logic similar to what's [used in `pgc-ngp-van`](https://github.com/politics-rewired/pgc-ngp-van/blob/master/src/task-list.ts#L44-L46) for the `VanOperationMode` resolver.

It may be easier to remove `VanOperationMode` entirely and only deal with the suffix. Client-side, both the text field and the operation mode dropdown would deal with the `apiKey` property. Thoughts on this @ben-pr-p?

src/api/external-system.ts
```diff
@@ -29,13 +35,19 @@ export interface ExternalSystemInput {
...
 
 export const schema = `
   enum ExternalSystemType {
     VAN
   }
 
+  enum VanOperationMode {
+    Voterfile
+    MyCampaign
+  }
```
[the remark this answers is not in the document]
ben-pr-p: I think it's good to have an operation mode resolver / graphql field that is just derived / stored from the `|0` or `|1` in the api key. Otherwise, there's only a chance they can get out of sync if that info is stored in two places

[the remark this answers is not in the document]
henryk1229: 👍  makes sense

src/server/api/external-system.ts
```diff
@@ -77,6 +77,14 @@ export const resolvers = {
         .reader("external_result_code")
         .where({ system_id: system.id });
       return formatPage(query, { after, first, primaryColumn: "created_at" });
+    },
+    operationMode: async (system: ExternalSystem) => {
+      const components = system.api_key_ref.split("|");
+      const operationMode = components[2];
```
bchrobot, line 83: I think this should be `1`?

```suggestion
      const operationMode = components[1];
```

bchrobot (changes requested):

src/server/api/external-system.ts
henryk1229: It looks like the org Id is pre-pended to the apiKey, so following the pattern on lines 39-40, maybe it should be:

`operationMode = components.length === 3 ? components[2] : components[1]`

bchrobot: ah, yep, this is what you want:

```suggestion
      const operationMode = components.length === 3 ? components[2] : components[1];
```

bchrobot (approved):
Looks good to me once conflicts are resolved and the PR body is filled in 👍

Issue #782, opened 2020-11-17
User email addresses can be duplicated


bchrobot:
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**

There are no database checks for email address uniqueness. This check is performed in application code and we have seen it fail. We know that in one instance this lead to an account lockout as the password reset operated on one user and the login checked against the other.

**To Reproduce**

Can not reliably reproduce.

**Expected behavior**

An email address may only be used one time.

**Screenshots**

N/A

**Additional context**

N/A

bchrobot:
This can be addressed in one of two ways:

1. Add a unique index:
    ```sql
    create unique index email_unique_idx on public.user (lower(email));
    ```
1. Switch to case-insensitive text and add a regular unique constraint:
    ```sql
    create extension if not exists citext;
    alter table public.user alter column email type citext;
    alter table public.user add constraint email_unique unique (email);
    ```

bchrobot:
This will also require writing a cleanup SQL script to deduplicate existing databases. Current business logic should be checked to see make sure we're leaving intact the user record that the login code runs against.

ben-pr-p:
For two accounts with the same email, we should:

- Keep the password with the most recently used account. The most efficient way to determine this given existing indexes is by most recently created assignment.
- Deprecate the old account by modifying the email address to add `+deprecatedviaduplicate` before the `@`
- Merge/union organization role memberships via `insert into select from` + an on conflict update user_id
- Delete organization memberships associated with the old user
- Reassign assignments (`update assignment set user_id = new_user_id where user_id = old_user_id`) for non-archived campaigns

PR #821, opened 2021-01-04
fix(user): unique email addresses with migration

ben-pr-p:
## Description

Changes user.email to citext and includes some data migration to make the constraint true. For details on the migration, see the comment in the issue: https://github.com/politics-rewired/Spoke/issues/782

## Motivation and Context

See https://github.com/politics-rewired/Spoke/issues/782

## How Has This Been Tested?

I've tested that the migration is syntactically correct and up/downable and doesn't mess up a clean database locally.

The accuracy of the data modification has been tested on a client's production database with rollbacks.

## Types of changes

## Checklist:

### DIFF
--- a/migrations/20210104095122_unique_user_email_addresses.js
+++ b/migrations/20210104095122_unique_user_email_addresses.js
@@ -1,0 +1,83 @@
+exports.up = function (knex) {
+  return knex.schema.raw(`
+    create extension if not exists citext;
+    alter table public.user alter column email type citext;
+
+    with 
+      duplicated_users as (
+        select email
+        from public.user
+        group by 1
+        having count(*) > 1
+      ),
+      user_assignment as (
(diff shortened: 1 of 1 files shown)


bchrobot (changes requested):
The deduplication statement looks good to me. I don't think it should be run automatically as part of the migration, however. I think that should live either in a comment in the migration and/or in the PR description (see note in #722).

ben-pr-p:
> The deduplication statement looks good to me. I don't think it should be run automatically as part of the migration, however. I think that should live either in a comment in the migration and/or in the PR description (see note in #722).

Why shouldn't it live in the migration? If it doesn't, we could have a bunch of failed migrations leading to failed init containers where we then have to go to each one and run it

bchrobot (approved):

Issue #780, opened 2020-11-17
Survey selected but script not sent


bchrobot:
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**

Texters have reported selecting a survey response and clicking send only to find later that only recording the survey response succeeded and the text itself never sent.

**To Reproduce**

Can not reliably reproduce.

**Expected behavior**

Send message actions should succeed or fail as a group: the actual sending of the message along with any data collection.

**Screenshots**

N/A

**Desktop (please complete the following information):**

N/A

**Additional context**

N/A

lh00000000:
what level of confirmed-sent-ness is desired here? would this require waiting for a delivery report?

bchrobot:
The issue is not with the `sendMessage` mutation itself. Rather that the client performs multiple GraphQL requests which may succeed/fail independently. In the case of this bug report, the `sendMessage` mutation request failed but the `updateQuestionResponses` mutation request succeeded.

A few of the options to address this are:
1. Wrap all "conversation action" operations in one request. The backend would then enforce these all succeeding/failing as a group. This could be done:
    1. with a single DB transaction. the downside is that sending a message (especially if switchboard is feeling slow) can take up to 30 seconds and holding transactions for the duration could quickly lead to connection pool exhaustion
    2. treating message sending as an internal saga of sorts with a more complicated rollback process for a failure in the vendor message sending stage
2. a client-side band-aid to better capture/surface errors in these request groups

PR #822, opened 2021-01-04
refactor(send-message): add general handle conversation

ben-pr-p:
## Description

This is a defensive fix to address issues likely responsible for https://github.com/politics-rewired/Spoke/issues/780 where some of `sendMessage`, `updateQuestionResponse`, etc. succeed and some do not.

It moves these operations to one mutation and runs them in a single transaction on the server.

## Motivation and Context

See https://github.com/politics-rewired/Spoke/issues/780.

## How Has This Been Tested?

I've tested each code pathway locally - send a message while updating a question response, while deleting a question response, while opting out a contact, and all of the above while skipping a reply as well.

## Types of changes

## Checklist:

### DIFF
--- a/src/api/schema.js
+++ b/src/api/schema.js
@@ -305,6 +305,14 @@ const rootSchema = `
     editCampaignContactMessageStatus(messageStatus: String!, campaignContactId:String!): CampaignContact,
     deleteQuestionResponses(interactionStepIds:[String], campaignContactId:String!): CampaignContact,
     updateQuestionResponses(questionResponses:[QuestionResponseInput], campaignContactId:String!): CampaignContact,
+    handleConversation(
+      campaignContactId: String!, 
+      message:MessageInput,
+      questionResponses: [QuestionResponseInput],
+      interactionStepIdsForDeletedQuestionResponses: [String],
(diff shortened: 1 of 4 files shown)


bchrobot (changes requested):
If you're already pulling the business logic up and out of the resolver, I think it makes sense to move it to separate files. `schema.js` is just so bloated.

Where that should live, I'm not sure. I haven't found many apollo-server directory structure examples that seem like a good fit.

ben-pr-p:
> If you're already pulling the business logic up and out of the resolver, I think it makes sense to move it to separate files. `schema.js` is just so bloated.
> 
> Where that should live, I'm not sure. I haven't found many apollo-server directory structure examples that seem like a good fit.

I don't think it makes sense to just move these functions to a separate file. Subsets of mutations as a whole can go to other files, but that can be a separate PR

ben-pr-p:
> If you're already pulling the business logic up and out of the resolver, I think it makes sense to move it to separate files. `schema.js` is just so bloated.
> 
> Where that should live, I'm not sure. I haven't found many apollo-server directory structure examples that seem like a good fit.

Later!

Issue #820, opened 2021-01-03
Saving teams doesn't work properly


aradox66:
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**
Saving new teams to a campaign doesn't work unless 'restrict to teams' is selected. Currently, to get the function to work reliably, you have to add teams, restrict to those teams and save, then refresh, unrestrict and save again.

If you save the teams without restricting, the teams will disappear on refresh.

**Expected behavior**
The teams should save without needing to restrict the campaign to those teams.

**Desktop (please complete the following information):**
Chrome: Version 87.0.4280.109 (Official Build) (64-bit)
Mozilla Firefox 83.0

ben-pr-p:
Thanks for the issue @aradox66 !

Adding to Q4 feature freeze as it meets the requirements and is reproducible.

aradox66:
Amazing, thanks so much.

PR #825, opened 2021-01-06
fix: remove mutation of props in CampaignTeamsForm

lh00000000:
previous code was imperatively mutating the teams array passed in as props

## Description
AdminCampaignEdit relies on comparison between props and server response to enable or disable the save button. When editing a campaign's teams, only toggling the non-teams-array related value (the flag for restricting texters to team) is the only way to allow this condition to be true, as the teams array (pulled from server response and passed as props) was being mutated on user input, so it was always in sync (regardless of database changes).

## Motivation and Context
this should fix this issue: https://github.com/politics-rewired/Spoke/issues/820

## How Has This Been Tested?
only tried locally

## Types of changes

## Checklist:

### DIFF
--- a/src/containers/AdminCampaignEdit/index.jsx
+++ b/src/containers/AdminCampaignEdit/index.jsx
@@ -1,4 +1,5 @@
 import gql from "graphql-tag";
+import { pick } from "lodash";
 import isEqual from "lodash/isEqual";
 import Avatar from "material-ui/Avatar";
 import { Card, CardActions, CardHeader, CardText } from "material-ui/Card";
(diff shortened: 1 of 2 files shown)


bchrobot (approved):
🚢

src/containers/AdminCampaignEdit/index.jsx
```diff
@@ -1,4 +1,5 @@
 import gql from "graphql-tag";
+import { pick } from "lodash";
```
ben-pr-p, line 2: We decided to do imports from the lodash sublibrary, right?

src/containers/AdminCampaignEdit/sections/CampaignTeamsForm.jsx
```diff
@@ -1,3 +1,4 @@
+import { differenceBy } from "lodash";
```
ben-pr-p, line 1: We decided to do imports from the lodash sublibraries, right?

bchrobot: Good catch. Addressed in 46db2b64270dd20dc845a33387c41b7bc2cbb42a

src/containers/AdminCampaignEdit/index.jsx
bchrobot: Good catch. Fixed in 46db2b64270dd20dc845a33387c41b7bc2cbb42a

ben-pr-p:
Closes https://github.com/politics-rewired/Spoke/issues/820

PR #850, opened 2021-01-07
feat(users): add superadmin option to role selection

### LINKED ISSUE #814: Add Superadmin option to role selection

There is no way for a superadmin to manage other superadmins.

A "Superadmin" option should be added to the role selection on the People page. This option should only be selectable by other superadmins and (maybe) should require a dialog confirmation.

A "makeSuperadmin" option should be added to invite payload to make the user who redeems that invite a superadmin.

A superadmin dashboard separate from existing Spoke mini-apps.

**Additional context**

N/A

henryk1229:
## Description

Closes #814, part of Q4 feature freeze.

## Motivation and Context

This allows superAdmins to manage other superAdmins by selecting the role within the People page. It's limited to other superAdmins within the role hierarchy, and selecting 'superadmin' prompts a confirmation dialog before running the mutation to change the role.

## How Has This Been Tested?

This has been tested locally.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:


ben-pr-p:
- [ ] Set `user` `is_superadmin = true` on set to superadmin
- [ ] Return super admin in org membership resolver
- [ ] Remove migration

ben-pr-p (changes requested):
We need to make sure that only `SUPERADMIN`'s can modify someone else's `SUPERADMIN` status on the server as well. 

Modifying https://github.com/politics-rewired/Spoke/pull/850/files#diff-6167f7d4b5d9664e676a5d148c2a2c316b56085d6ac43620eaab1aeb07e02e96R982 is probably the best place to do it

src/server/api/schema.js
```diff
@@ -989,6 +989,22 @@ const rootMutations = {
...
+        .returning("*");
+
+      // handle Superadmin role selection
+      if (role === "SUPERADMIN") {
+        user.update({ is_superadmin: true });
+      }
+      // handle Superadmin role de-select
+      if (user.is_superadmin && role !== "SUPERADMIN") {
+        user.update({ is_superadmin: false });
+      }
```
bchrobot, line 1006: There are a few issues with the update here.

1. This user update query is never actually run. If you look below, there's a `const [orgMembership] = await updateQuery;` that actually kicks off the `updateQuery` query. In general I think it's helpful to include "query" in the variable name as a reminder that it's a query being built, not an ORM object.
2. To capture the edge cases introduced by is_superadmin living on a different table I think we always want to make two update queries:
    1. `user.update({ is_superadmin: role === UserRoleType.SUPERADMIN });`
    2. `if (role) updateQuery.update({ role: role === UserRoleType.SUPERADMIN ? UserRoleType.OWNER : role });`

So for this block:

```suggestion
      const userUpdateQuery = r
        .knex("user")
        .where({ id: membership.user_id });
      if (role) userUpdateQuery.update({ is_superadmin: role === UserRoleType.SUPERADMIN });
```

and then updating below to be:

```ts
const [orgMembership, _] = await Promise.all([updateQuery, userUpdateQuery]);
```

(we don't care about returning a result from `userUpdateQuery` -- that'll be done by the `Membership.role` resolver)

bchrobot (changes requested):
There shouldn't be changes to `yarn.lock` (unless those changes weren't committed alongside their associated `package.json` changes).

Try doing:

```sh
# Checkout just yarn.lock from the master branch
git checkout master -- yarn.lock

yarn install

git diff
```

ben-pr-p:
> There shouldn't be changes to `yarn.lock` (unless those changes weren't committed alongside their associated `package.json` changes).
> 
> Try doing:
> 
> ```shell
> # Checkout just yarn.lock from the master branch
> git checkout master -- yarn.lock
> 
> yarn install
> 
> git diff
> ```

I'm also getting changes to `yarn.lock` when installing locally, so somewhere along the line someone forgot to include those. If we don't want those to be included in this PR, @bchrobot would you mind committing those to master and rebasing this?

henryk1229:
I've got the changes from the review, but for some reason the UI still isn't reflecting the change after the editOrganizationMembership mutation runs if you assign or un-assign a user as a Superadmin, or change a user's Auto Approve level - thoughts?

bchrobot (approved):

PR #857, opened 2021-01-12
feat: per organization message limits

ben-pr-p:
## Description

Optionally restrict the amount of outbound messages an organization can send per month.

## Motivation and Context

User request for cost controls

## How Has This Been Tested?

I checked locally:
- [x] You can't send initials when you're over the limit
- [x] You can send initials when you're under the limit
- [x] You can send replies when you're over the limit
- [x] The backfill function is correct
- [x] The cronjob runs and produces the right behavior

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/migrations/20210112103258_monthly-org-message-limits.js
+++ b/migrations/20210112103258_monthly-org-message-limits.js
@@ -3,7 +3,7 @@ exports.up = (knex) => {
     alter table organization add column monthly_message_limit bigint;
 
     create table monthly_organization_message_usages (
-      month integer,
+      month date,
       organization_id integer references organization (id),
       sent_message_count bigint,
       primary key (organization_id, month)
@@ -13,12 +13,12 @@ exports.up = (knex) => {
       insert into monthly_organization_message_usages (organization_id, month, sent_message_count)
       select 
         campaign.organization_id, 
-        extract('month' from now()) as month,
+        date_trunc('month', now()) as month,
         count(*) as sent_message_count
       from message
       join campaign_contact on campaign_contact.id = message.campaign_contact_id
       join campaign on campaign.id = campaign_contact.campaign_id
-      where message.created_at > date_trunc('month', now())
+      where message.created_at >= date_trunc('month', now())
         and message.is_from_contact = false
       group by 1, 2
       on conflict (organization_id, month)
@@ -32,6 +32,6 @@ exports.up = (knex) => {
 exports.down = (knex) => {
   return knex.schema.raw(`
     alter table organization drop column monthly_message_limit;
-    drop table organizaton_usages;
+    drop table monthly_organization_message_usages;
   `);
 };
--- a/migrations/20210112103258_monthly-org-message-limits.js
+++ b/migrations/20210112103258_monthly-org-message-limits.js
@@ -1,0 +1,37 @@
+exports.up = (knex) => {
+  return knex.schema.raw(`
+    alter table organization add column monthly_message_limit bigint;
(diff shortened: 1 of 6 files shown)


src/server/tasks/update-org-message-usage.ts
```diff
@@ -0,0 +1,36 @@
+import { JobHelpers, Task } from "pg-compose";
+
+import { config } from "../../config";
+
+const updateOrgMessageUsage: Task = async (
+  payload: any,
+  helpers: JobHelpers
+) => {
+  if (config.ENABLE_MONTHLY_ORG_MESSAGE_LIMITS) {
```
bchrobot, line 9: nit: I don't think we want a duplicate check here. There's already a check on before scheduling the job and, who knows, we may want to kick off a backfill job on an instance with the envvar disabled

src/server/tasks/update-org-message-usage.ts
```diff
@@ -0,0 +1,36 @@
...
+        insert into monthly_organization_message_usages (organization_id, month, sent_message_count)
+        select 
+          campaign.organization_id, 
+          extract('month' from now()) as month,
+          count(*) as sent_message_count
+        from message
+        join campaign_contact on campaign_contact.id = message.campaign_contact_id
+        join campaign on campaign.id = campaign_contact.campaign_id
+        where message.created_at > date_trunc('minute', $1) - interval '5 minutes'
+          and message.created_at <= date_trunc('minute', $1)
```
bchrobot, line 21: I think better to have the equality on the lower bound
```suggestion
        where message.created_at >= date_trunc('minute', $1) - interval '5 minutes'
          and message.created_at < date_trunc('minute', $1)
```

bchrobot (changes requested):

migrations/20210112103258_monthly-org-message-limits.js
```diff
@@ -0,0 +1,37 @@
...
+      insert into monthly_organization_message_usages (organization_id, month, sent_message_count)
+      select 
+        campaign.organization_id, 
+        extract('month' from now()) as month,
+        count(*) as sent_message_count
+      from message
+      join campaign_contact on campaign_contact.id = message.campaign_contact_id
+      join campaign on campaign.id = campaign_contact.campaign_id
+      where message.created_at > date_trunc('month', now())
+        and message.is_from_contact = false
```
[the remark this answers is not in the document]
ben-pr-p: Changing to `>=` but the upper bound doesn't make sense here since it'd just be filtering out future messages

src/server/tasks/update-org-message-usage.ts
ben-pr-p: Keeping it because by default removing a job from graphile-scheduler doesn't deschedule it from the `schedules` table in the database, so this is necessary to prevent it from actually running

bchrobot: Ah, that makes sense 👍

migrations/20210112103258_monthly-org-message-limits.js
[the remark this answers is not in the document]
bchrobot: I got the `insert` arguments confuse with the signature and read the function signature as `backfill_month_organization_message_usages(month integer)`, in which case it could be used for historical backfilling and would need an upper bound. Fine to not think about historical backfilling though I think

bchrobot (approved):

Issue #829, opened 2021-01-06
Display character count in Texter view


kohlee:
**Is your feature request related to a problem? Please describe.**
Texters sometimes send messages that exceed desired or sendable character counts. 

**Describe the solution you'd like**

1. Messages drafted in Texter view should display a character count, as is done in the Interactions tab.
1. Messages longer than 1600 characters should display a warning that a message of that length is unsendable.

**Describe alternatives you've considered**
Messages drafted in Texter view could display "Message too long" rather than the exact character count after a set number of characters. This cutoff could be decided by the Admin in Settings or statically set on the back end.

hblumberg:
Hi, @kohlee! Seems like this might be a duplicate of / related to #438? Hoping to put up a PR fix for that one soon 🤞

bchrobot:
@hblumberg @kohlee Yeah these are the same underlying issue. The two things in this issue not in 438 are:

1. Adding a character counter to the composition field in the admin Message Review (which always gets forgotten about)
2. Adding an explicit warning about maximum SMS length (438 just says to prevent text entry, but the user or admin probably wants to know _why_ they can't keep typing)

Issue #711, opened 2020-08-18
Incremental UI improvement to make mobile texting possible


arena:
The main issue that makes the mobile experience so painful is the green arrow that overlaps the text field. I propose an incremental fix of removing that green arrow. Then, add a row to the bottom with "SEND RESPONSE" and SKIP 

Note: Send Response has a long label to make the button bigger and more of an easy default target.

Mockup
https://www.figma.com/file/kBYElg9lNTrzkn9Oe4RlJL/PoliticsRewired-Spoke-Incremental-Fix?node-id=0%3A1

arena:
screenshot of figma (as of when I'm posting this)
![image](https://user-images.githubusercontent.com/411070/90572361-b18ede00-e168-11ea-9640-84f19d05b147.png)

arena:
![image](https://user-images.githubusercontent.com/411070/90574976-2d8c2480-e16f-11ea-8e85-b92599d34f82.png)

kcarrandale:
One issue we've seen with buttons is that is volunteers don't always realize the tags are hidden behind the three-dot menu button (and training can only help so much). Unless there's a plan to add features down the line, could this be turned into a button on its own (like "Tags" to save on space)? It would probably fit reasonably well in one of the mockups you propose.

jbradach:
> One issue we've seen with buttons is that is volunteers don't always realize the tags are hidden behind the three-dot menu button (and training can only help so much). Unless there's a plan to add features down the line, could this be turned into a button on its own (like "Tags" to save on space)? It would probably fit reasonably well in one of the mockups you propose.

Tags in their current form aren't very useful because volunteers don't really look for them. Unless there's a big overhaul in the interface, I don't anticipate any of our members using them for most of their campaigns. 

Really like having opt-out away from other commonly used buttons. I just hate that we 1) people replying "STOP" aren't automatically filtered out like on other platforms requiring more volunteer time to opt those folks out and 2) by default a message confirming the opt-out is sent to the person which just doesn't make sense when they don't want to hear from us so we're wasting money on an extra segment.

arena:
@jbradach seems like we should write a separate issue around response (and not have to respond) to STOP replies

Issue #438, opened 2020-01-09
Add character/segment count to message composition field


bchrobot:
**Is your feature request related to a problem? Please describe.**

There are two issues:

1. Organizations may accidentally send, and be billed for, many more segments than anticipated
2. Texters may spend time crafting a very length response only to have the message fail because it is too long.

**Describe the solution you'd like**

The message composition field should:

- Display current character count
- Display current segment count
- Prevent text entry when maximum character count has been reached
- (Optionally) display encoding (GSM vs UCS2) required

**Resources**
- https://chadselph.github.io/smssplit/
- https://www.npmjs.com/package/gsm

hblumberg:
Hi, @bchrobot! I'd be happy to try to take this one on (: 

Did you have a specific design in mind? If not, what do you think of something like this?
![Screen Shot 2021-01-02 at 9 34 30 PM](https://user-images.githubusercontent.com/8495791/103470635-a36de280-4d42-11eb-9205-4f3c98ffd004.png)
![Screen Shot 2021-01-02 at 9 38 03 PM](https://user-images.githubusercontent.com/8495791/103470660-dc0dbc00-4d42-11eb-82f8-22b6b5f38dac.png)
![Screen Shot 2021-01-02 at 9 35 03 PM](https://user-images.githubusercontent.com/8495791/103470661-ddd77f80-4d42-11eb-9fd6-eaeedfca5698.png)

bchrobot:
Hey @ hblumberg that would be great, thank you!

PR #859, opened 2021-01-18
fix(sendbuttonarrow): pin the send arrow to the top of the message field

hblumberg:
## Description

Before this change, the send arrow was positioned absolutely with respect to the entire screen, and it was "pinned" to the bottom right corner.

With this change, the send arrow is positioned absolutely with respect to the message field, and it is "pinned" to the top right corner.

This change also hides the send arrow when the screen width is greater than 450px. The existing CSS suggests this was the desired behavior, but it wasn't working as expected.

## Motivation and Context

While I was working on #438 / #829, I realized that adding the character count below the text field would require moving the arrow up by an arbitrary amount (the height of the new component). It felt off to have the `SendButtonArrow` component so dependent on its sibling components.

I also noticed #711, which suggests the incremental change of moving the send arrow out of the text field.

This change makes `SendButtonArrow` less dependent on its siblings (specifically the bottom action bar) and moves the arrow out of the text field, to the top right.

## How Has This Been Tested?

I tested this change in dev on Chrome and Safari with different screen widths.

## Screenshots:

before
after

## Types of changes

## Documentation Changes

n/a

## Checklist:

### DIFF
--- a/src/components/SendButtonArrow.jsx
+++ b/src/components/SendButtonArrow.jsx
@@ -4,12 +4,14 @@ import NavigationArrowForward from "material-ui/svg-icons/navigation/arrow-forwa
 import PropTypes from "prop-types";
 import React, { Component } from "react";
 
-// This is because the Toolbar from material-ui seems to only apply the correct margins if the
-// immediate child is a Button or other type it recognizes. Can get rid of this if we remove material-ui
+const INTERNAL_ARROW_PADDING = 11;
+
 const styles = StyleSheet.create({
   container: {
-    display: "inline-block",
-    marginLeft: 20
+    position: "absolute",
+    right: -INTERNAL_ARROW_PADDING,
+    top: -INTERNAL_ARROW_PADDING,
+    padding: 0
   },
   icon: {
     color: "rgb(83, 180, 119)",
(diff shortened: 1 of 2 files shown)


src/components/SendButtonArrow.jsx
```diff
@@ -4,29 +4,28 @@ import NavigationArrowForward from "material-ui/svg-icons/navigation/arrow-forwa
...
   },
   arrowButton: {
-    "@media(minWidth: 450px)": {
+    "@media(min-width: 450px)": {
       display: "none !important"
     },
-    "@media(maxWidth: 450px)": {
+    "@media(max-width: 450px)": {
       display: "block !important"
     },
```
hblumberg, line 27: If the send arrow should be visible on larger screen sizes (as it was before this change), I can remove this for the following result.

after-wo-media

bchrobot: Nope, it should be hidden on larger screens! I think there was just originally some confusion between the camel case React expects for CSS rule names and the standard hyphen case CSS uses.

bchrobot (approved):
Thanks!

PR #862, opened 2021-01-20
refactor: manage pagination result with apollo

bchrobot:
## Description

Use Apollo Client caching for pagination.

## Motivation and Context

Managing pagination state manually means components will not get updates from the Apollo Client cache (e.g. mutation result). This uses the `<Query>` component from Apollo Client to push pagination state management up to Apollo.

## How Has This Been Tested?

This has been tested locally.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "hot-shots": "^6.3.0",
     "humps": "^1.1.0",
     "iconv-lite": "^0.5.1",
+    "immer": "^8.0.1",
     "isomorphic-fetch": "^2.2.1",
     "knex": "^0.21.1",
     "lodash": "^4.13.1",
--- a/src/components/InfiniteRelayList.tsx
+++ b/src/components/InfiniteRelayList.tsx
@@ -1,4 +1,5 @@
 import { FetchPolicy } from "apollo-client";
+import produce, { Draft } from "immer";
 import React from "react";
 import { Query } from "react-apollo";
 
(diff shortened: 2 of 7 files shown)


lh00000000:
just looked at the diffs, so i might have missed something, but how does this handle when a graphql query+variables pair will return a different response because the database just handled some update mutation send by some other component? i.e. what `dbLastUpdatedAt` was there for before

ben-pr-p (approved):
Does this fix an active bug(s)?

lh00000000:
> > just looked at the diffs, so i might have missed something, but how does this handle when a graphql query+variables pair will return a different response because the database just handled some update mutation send by some other component? i.e. what `dbLastUpdatedAt` was there for before
> 
> It becomes the mutation's job to update the cache accordingly. For updating existing objects, Apollo does that automatically based on the mutation's response. In this case, the `editOrganizationMembership` mutation returns the updated OrganizationMembership object. Apollo Client sees this new version of the object, uniquely identified by its `__typename` and `id`, and updates all occurrences of that object in other queries.
> 
> For create/delete mutations, the pagination result is manipulated directly by passing the query an `update()` option. This isn't widely used in Spoke but is the preferred way of doing things. The [`createConfig`/`deleteConfig` mutations](https://github.com/politics-rewired/Spoke/blob/bc04b78c906ccbebe3a80ed0e76f4cc5228c6415/src/components/SyncConfigurationModal/index.tsx#L144-L185) for configuring VAN sync are one example.

oh sorry, i missed your reply. i haven't gotten a chance to look at the `update()` option yet but sounds good to me !

PR #868, opened 2021-01-22
fix(vanoperationmode): remove operation mode key from external system mutations

henryk1229:
## Description

Per @ben-pr-p, fix for an error introduced in #813, it removes writing to the non-existent operation_mode column in the database. Van Operation Mode is handled by the External System Resolver, but the write wasn't removed from the mutation after the change was implemented in the original PR.

## Motivation and Context

Bug fix

## How Has This Been Tested?

This has been tested locally.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -3390,8 +3390,7 @@ const rootMutations = {
           type: externalSystem.type.toLowerCase(),
           organization_id: parseInt(organizationId, 10),
           username: externalSystem.username,
-          api_key_ref: apiKeyRef,
-          operation_mode: externalSystem.operationMode
+          api_key_ref: apiKeyRef
         })
         .returning("*");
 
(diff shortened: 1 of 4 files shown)


bchrobot (approved):

src/server/api/schema.js
```diff
@@ -3411,64 +3408,65 @@ const rootMutations = {
...
 
-        if (operationMode === VanOperationMode.VOTERFILE) {
-          apiKeyRef = apiKeyRef.concat("|0");
-        }
+      const apiKeyAppendage =
+        externalSystem.operationMode === VanOperationMode.MYCAMPAIGN
+          ? "|1"
+          : "|0";
+
+      const truncatedKey = `${externalSystem.apiKey.slice(0, 5)}********`;
```
bchrobot, line 3444: Max gave some feedback a while back that we might want to truncate the beginning and leave the last 4 in plaintext because the last 4 are what VAN uses to uniquely identify keys. Is that something we could drop in as part of this PR?

bchrobot (approved):

src/server/api/schema.js
ben-pr-p: I'll make an issue - I think the QA required means we can't get that out today

PR #889, opened 2021-02-12
build: drop old proxy deps

bchrobot:
## Description

Update dependencies to drop downstream dependencies on old versions of `http-proxy-agent` and `socks-proxy-agent`.

This also drops support for Mailgun as a) we are not using it and b) there are no stable versions of a node SDK.

## Motivation and Context

The most common stack traces of the `this.socket.cork is not a function` error come from either `connect-datadog-graphql` or the express router but both are [via the `finalhandler` library](https://github.com/pillarjs/finalhandler/blob/master/index.js#L297). Looking at the final line of the trace, [`ServerResponse.end` in v12.20.2 of Node](https://github.com/nodejs/node/blob/v12.20.2/lib/_http_outgoing.js#L751-L753), we find our call to cork:

```js
  if (this.socket) {
    this.socket.cork();
  }
```

Looking at [`http-proxy-agent@1.0.0`'s `connect()`](https://github.com/TooTallNate/node-http-proxy-agent/blob/1.0.0/http-proxy-agent.js#L85-L90) we find:

```js
function connect (req, opts, fn) {

  // ...

  // create a socket connection to the proxy server
  var socket;
  if (this.secureProxy) {
    socket = tls.connect(proxy);
  } else {
    socket = net.connect(proxy);
  }

  // ...

  fn(null, socket);
}
```

Which starts to paint a picture of an older version of `http-proxy-agent` passing an older socket without `cork()` to `ServerResponse`. I have no idea where that substitution be happening, however.

Stack traces:

<details>
<summary><code>connect-datadog-graphql</code></summary>

```
TypeError: this.socket.cork is not a function
    at ServerResponse.end (_http_outgoing.js:752:17)
    at ServerResponse.res.end (/usr/Spoke/node_modules/connect-datadog-graphql/lib/index.js:50:8)
    at Array.write (/usr/Spoke/node_modules/finalhandler/index.js:297:9)
    at listener (/usr/Spoke/node_modules/on-finished/index.js:169:15)
    at onFinish (/usr/Spoke/node_modules/on-finished/index.js:100:5)
    at callback (/usr/Spoke/node_modules/ee-first/index.js:55:10)
    at IncomingMessage.onevent (/usr/Spoke/node_modules/ee-first/index.js:93:5)
    at IncomingMessage.emit (events.js:314:20)
    at IncomingMessage.EventEmitter.emit (domain.js:483:12)
    at endReadableNT (_stream_readable.js:1241:12)
[...]

PR #890, opened 2021-02-13
refactor: use lightship instead of terminus

bchrobot:
## Description

Use Lightship instead of Terminus.

## Motivation and Context

Unconvinced that #889 would fully solve the socket.cork problem, I continue working on the Node 14 upgrade. I was able to reproduce the `this.socket.cork` error locally on every request for `/`. Initially I tried upgrading other socket-based libraries but that did not solve the problem. I revisited source code for Node requests and the packages in the stack trace an decided to try slimming down the express app.

Removing Terminus and just using `app.listen(...)` solved the problem. I believe we had tried removing Terminus previously and it did not solve the issue but that may have been a mistake in the rapid deployment iterations we were doing. Looking at the Terminus source code there are at least two places that the Express `server` is modified: [`decorateWithSignalHandler()`](https://github.com/godaddy/terminus/blob/master/lib/terminus.js#L103-L105) and [`decorateWithHealthCheck()`](https://github.com/godaddy/terminus/blob/master/lib/terminus.js#L94-L97). In `decorateWithSignalHandler()` the server is monkey patched with [`stoppable`](https://www.npmjs.com/package/stoppable), a package that hasn't been updated in 2 years.

Lightship:

1. works with Node 14 with no issues,
2. Is built specifically with Kubernetes in mind, and
3. just has a better design overall where health checks are managed as a separate service.

`hot-shots` was updated to fix typing issues.

## How Has This Been Tested?

General Spoke functionality has been tested locally.

The update to StatsD cannot be tested locally. The [API](https://github.com/brightcove/hot-shots#check) is compatible with [`connect-datadog-graphql`'s usage](https://github.com/politics-rewired/node-connect-datadog-graphql/blob/master/lib/index.js#L97-L101).

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:


ben-pr-p (approved):
👍 

https://github.com/politics-rewired/Spoke/pull/890/files#diff-df08de68ff1ed17a666c10aaeb765a0266b8b7370ee3956634506706dc97c4afL207 seems like the key line to me where a monkey patch could be occurring

PR #894, opened 2021-02-19
build: update react to v16

bchrobot:
## Description

This updates React to version 16.

## Motivation and Context

Stability, security, and same development platform across projects.

Depends on #897 and #898

## How Has This Been Tested?

This has been tested locally.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:


src/components/AuthzProvider.jsx
```diff
@@ -63,8 +63,11 @@ AuthzProvider.propTypes = {
   organizationId: PropTypes.string.isRequired
 };
 
-export const withAuthzContext = (Component) => (props) => (
-  <AuthzContext.Consumer>
-    {(adminPerms) => <Component {...props} adminPerms={adminPerms} />}
-  </AuthzContext.Consumer>
-);
+export const withAuthzContext = (Component) =>
+  function WithAuthzContext(props) {
```
ben-pr-p, line 67: The old version is in line with our lint rules, right?

ben-pr-p (approved):
Looks good!

Maybe we want to just write down what the approach to migrating off of `UNSAFE_componentWillMount` or other `UNSAFE_` render methods will be, and when it will be appropriate in the course of modifying a component to do that? I'm not 100% why they are unsafe, either, but I think just knowing when to keep using those or not will be good

src/components/AuthzProvider.jsx
bchrobot: It's never complained before. I think there's a `reactVersion: detect` somewhere, possibly internal to the ESLint plugin, that just started catching it?

PR #926, opened 2021-03-17
fix(manual-assignment): remove unused expensive query

ben-pr-p:
## Description

Some clients with larger `campaign_contact` tables reporting manual assignment failing - upon investigation, this query was responsible, sometimes taking over 30 seconds and causing the whole transaction to abort. 

The query could be optimized as:
```sql
 select assignment.id as id 
from assignment 
where not exists (
  select 1
  from campaign_contact 
  where campaign_contact.assignment_id = assignment.id
    and campaign_contact.archived = false
)
and assignment.campaign_id =  $1;
```
but it wasn't being used anyways, so we can just remove it!

The query was originally part of a check to delete assignments which no longer had contacts associated with them, but we don't do that anymore since:
a) We might just end up recreating them in the future, and
b) It violated opt out foreign key constraints

## How Has This Been Tested?

This has not been tested

## Types of changes

## Documentation Changes

No

## Checklist:

- [] My change requires a change to the documentation.

### DIFF
--- a/src/workers/jobs/index.js
+++ b/src/workers/jobs/index.js
@@ -997,27 +997,6 @@ export async function assignTexters(job) {
         })
       );
 
-      // TODO: re-enable once dynamic assignment is fixed (#548)
-      // const isDynamic = campaign.use_dynamic_assignment;
-      const isDynamic = false;
-
-      // dynamic assignments, having zero initially is ok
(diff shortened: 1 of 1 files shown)


bchrobot (approved):
buhbyeeee unused query

PR #953, opened 2021-05-06
fix(van): fix sync for contacts with both opt out and responses

bchrobot:
## Description

Fix edge case where canvass result with both result code _and_ responses is sent to VAN by recording responses and opt-outs separately.

## Motivation and Context

This was prompted by a customer ticket regarding sync errors. Inspecting the Spoke logs showed a few 400 responses from VAN:

```json
{
  "errors": [
    {
      "code": "INVALID_PARAMETER",
      "text": "'resultCodeId' must be null when responses are specified.",
      "properties": [
        "resultCodeId"
      ]
    }
  ]
}
```

The only code path that allows a `resultCodeId` _and_ `responses` is if the contact is opted out. Spot checking the `campaign_contact` records for these failed syncs confirmed this hypothesis.

## How Has This Been Tested?

See changes to `sync-campaign-contact-to-van.spec.ts`.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/tasks/sync-campaign-contact-to-van.spec.ts
+++ b/src/server/tasks/sync-campaign-contact-to-van.spec.ts
@@ -150,8 +150,9 @@ describe("formatCanvassResponsePayload", () => {
       canvassedResultCode,
       optOutResultCode
     });
-    expect(canvassResponse.responses).toBeNull();
-    expect(canvassResponse.resultCodeId).toBe(optOutResultCode);
+    expect(canvassResponse).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


ben-pr-p:
Hm, this is actually kind of a bummer. I think the question responses recorded with an opt out are useful data. Do we want to maybe send over 2 events, the first for the responses, and the second for the opt out?

ben-pr-p (approved):
Gotcha, I misread some of the code there. Looks good!

PR #959, opened 2021-05-12
perf: consolidate texter assignment query

bchrobot:
## Description

Consolidate the assignment steps to one query and add tests.

## Motivation and Context

The current assignment queries are slow and untested.

## How Has This Been Tested?

See `assign-texters.spec.ts`.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/server/tasks/assign-texters.ts
+++ b/src/server/tasks/assign-texters.ts
@@ -1,3 +1,5 @@
+import { Pool, PoolClient } from "pg";
+
 import { DateTime } from "../../lib/datetime";
 import { CampaignRecord } from "../api/types";
 import { Notifications, sendUserNotification } from "../notifications";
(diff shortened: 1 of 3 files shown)


src/server/tasks/assign-texters.ts
```diff
@@ -11,6 +13,221 @@ interface Texter {
...
+  } = options;
+  await client.query(
+    `
+      update campaign_contact
+      set assignment_id = null
+      where
+        campaign_id = $1
+        and archived = ${isArchived}
+        and assignment_id is not null
+        and (cardinality($2::integer[]) = 0 or assignment_id <> ANY($2))
```
ben-pr-p, line 82: If `assignmentIds == []`, won't this un-assign every contact on the campaign?

src/server/tasks/assign-texters.ts
```diff
@@ -11,6 +13,221 @@ interface Texter {
...
+      with raw_assignments as (
+        select * from unnest($1::integer[], $2::integer[]) as t (assignment_id, desired_count)
+      ),
+      assignment_counts as (
+        select
+          assignment_id,
+          generate_series(1, desired_count - (
+            select count(*) from campaign_contact
+            where campaign_contact.assignment_id = raw_assignments.assignment_id
+          ))
```
ben-pr-p, line 166: `generate_series` in select statement with parameters! Very nice

src/server/tasks/assign-texters.ts
```diff
@@ -11,6 +13,221 @@ interface Texter {
...
+      assignable_contacts as (
+        select
+          row_number() over () as row,
+          id as campaign_contact_id
+          from campaign_contact
+        where
+          campaign_id = $3
+          and archived = ${isArchived}
+          and assignment_id is null
+          and message_status = 'needsMessage'
```
ben-pr-p, line 184: Should we be able to assign needs reply too?

I think we should preferably assign `needsMessage` and `needsResponse`, but using this section you should be able to assign all contacts on the campaign, right?

Also, performance wise this query would be improved with a limit on this stage in the CTE

ben-pr-p (commented):
Overall, looks like it should be a lot faster!

src/server/tasks/assign-texters.ts
bchrobot: That would change the function and behavior of the section and maybe should be a separate PR? I think that probably is the right direction to head in but I also think it's worth getting some product input (👋 @kohlee) first.

src/server/tasks/assign-texters.ts
bchrobot: Yep! And that is the intended behavior: `assignmentIds` is the list of assignments to keep. It has to be a separate cardinality check because:

```sql
assignment_id <> ANY('{}')
```

returns `false`.

I decided against a separate query altogether with a JS `if (assignmentTargets.length === 0)` check because the queries would be pretty similar with the `ignoreAfterDate` where clause.

src/server/tasks/assign-texters.ts
bchrobot: No join with the series required! I discovered a couple new SQL things with this refactor including that `for update` can't be used with a window function (`row_number()`) in this case.

src/server/tasks/assign-texters.ts
bchrobot: Moved to https://github.com/politics-rewired/Spoke/issues/964

PR #980, opened 2021-06-11
perf: fix sequential scan on campaign_contact

bchrobot:
## Description

Dramatically improve performance of `assignPayloads` query for clients with a large number of `campaign_contact` rows.

## Motivation and Context

There is only a partial index on `campaign_contact.assignment_id` so the `assignment_counts` CTE was doing a full scan
on `campaign_contact` for that stage. Adding the `campaign_id` condition turns that into an index scan
with `assignment_id` as a filter condition.

It could be further improved by limiting the use of the Texters section to unarchived campaigns: https://explain.depesz.com/s/fvsH

The `ensureAssignments` query being blocked by the `assignPayloads` query can likely be explained by:

1. User saves Texter section
2. `assignPayloads` takes a long time to run
3. User cancels the job; `assignPayloads` query continues to run in the background (see #981)
4. User saves Texters section again
5. Second job's `ensureAssignments` gets stuck on the first job's `assignPayloads`

## How Has This Been Tested?

Old explain: https://explain.depesz.com/s/uGxg
New explain: https://explain.depesz.com/s/iThU

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/tasks/assign-texters.ts
+++ b/src/server/tasks/assign-texters.ts
@@ -158,7 +158,9 @@ export const assignPayloads = async (options: AssignPayloadsOptions) => {
           assignment_id,
           generate_series(1, desired_count - (
             select count(*) from campaign_contact
-            where campaign_contact.assignment_id = raw_assignments.assignment_id
+            where
+              campaign_id = $3
+              and campaign_contact.assignment_id = raw_assignments.assignment_id
           ))
         from raw_assignments
       ),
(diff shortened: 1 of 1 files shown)


ben-pr-p (changes requested):
We should also add the partial condition (archived = false), and then let the planner decide between the partial assignment_id index (when the campaign is not archived) and the campaign_id index when the campaign is archived.

Should be a lot faster!

ben-pr-p (approved):
Let's roll this out asap?

PR #1011, opened 2021-09-02
fix: add missing primary keys

bchrobot:
## Description

Ensure every table has a primary key or a replica identity set.

## Motivation and Context

Clickhouse's MaterializedPostgreSQL engine requires that every synced table have a primary key or a replica identity set.

Replica identity seemed like the easiest approach for `messaging_service_stick` after attempting the following:

```sql
alter table public.messaging_service_stick
      add constraint messaging_service_stick_pkey primary key using index messaging_service_stick_cell_organization_unique_constraint;
```

but getting error:

```
   - index "messaging_service_stick_cell_organization_unique_constraint" is already associated with a constraint
```

## How Has This Been Tested?

Tables without primary keys were identified using:

```sql
select tab.table_schema,
       tab.table_name
from information_schema.tables tab
left join information_schema.table_constraints tco
          on tab.table_schema = tco.table_schema
          and tab.table_name = tco.table_name
          and tco.constraint_type = 'PRIMARY KEY'
where tab.table_type = 'BASE TABLE'
      and tab.table_schema not in ('pg_catalog', 'information_schema')
      and tab.table_schema = 'public'
      and tco.constraint_name is null
order by table_schema,
         table_name;
```

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/migrations/20210902181911_add-missing-primary-keys.js
+++ b/migrations/20210902181911_add-missing-primary-keys.js
@@ -1,0 +1,41 @@
+exports.up = function up(knex) {
+  return knex.schema.raw(`
+    alter table public.campaign_team
+      add column id serial primary key;
+
+    alter table public.user_team
+      add column id serial primary key;
+
+    alter table public.team_escalation_tags
+      add column id serial primary key;
+
+    -- We cannot turn existing unique index into primary key index without recreating the index
+    -- Just set the replica identity instead
+    alter table public.messaging_service_stick
+      alter column organization_id set not null,
+      alter column messaging_service_sid set not null,
+      alter column cell set not null;
+    alter table public.messaging_service_stick
+      replica identity using index messaging_service_stick_cell_organization_unique_constraint;
+  `);
+};
+
+exports.down = function down(knex) {
+  return knex.schema.raw(`
+    alter table public.campaign_team
+      drop column id;
+
+    alter table public.user_team
+      drop column id;
+
+    alter table public.team_escalation_tags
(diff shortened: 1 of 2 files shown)


migrations/20210902181911_add-missing-primary-keys.js
```diff
@@ -0,0 +1,41 @@
+exports.up = function up(knex) {
+  return knex.schema.raw(`
+    alter table public.campaign_team
+      add column id serial primary key;
+
+    alter table public.user_team
+      add column id serial primary key;
```
ben-pr-p, line 7: We can use (team_id, user_id) as the primary key

ben-pr-p:
> I think sticking with the id serial primary key convention is best

I think compound primary keys are good! It's equivalent to having multiple not null unique columns, and compound foreign keys (where another table will have a team_id and campaign_id for example) can be nice as well - you don't have to lookup to find an extra ID to make the foreign key work. Here it doesn't really matter, but I don't think we should be bound to id serial primary keys in general

> I didn't initially for messaging_service_stick thinking its write throughput is high enough that the existing unique index was enough

I have a slight preference for compound primary keys over id serials that isn't based on a performance perspective. `bigints` are small and fast and although it has a high write volume, it's all done during campaign upload time and not per send, which is what I might worry about.

That being said, I don't think it's worth adding because I don't think it's worth backfilling - that will be a very long running migration that substantially complicates the complexity of this rollout

PR #1047, opened 2022-01-20
fix: fix create campaign payload

bchrobot:
## Description

Fix payload used for `createCampaign` mutation.

## Motivation and Context

Addresses regression introduced in #1041.

## How Has This Been Tested?

This has been tested locally. Also see `campaign.spec.ts` changes.

I have also verified that GraphQL type `InteractionStepInput` is not used anywhere outside of `CampaignInput` which in turn is not used anywhere but `createCampaign` and `editCampaign` mutations.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:


ben-pr-p (commented):
Is the issue here that #1041 make id required? The fix and the bug look different to me, just making sure I understand

ben-pr-p:
In general it doesn't make sense to me to have a dummy id required for a create mutation (versus it not being required), and so I think a preferred fix would be reverting / fixing #1041 

I think this is fine to merge too, though

ajohn25 (approved):

PR #1064, opened 2022-02-08
build: generate types for pg relations

bchrobot:
## Description

Add generated TypeScript interfaces for Postgres relations.

## Motivation and Context

Consistent typing of database records that don't need to be maintained manually.

## How Has This Been Tested?

Reviewing the generated file.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

---
  - https://app.asana.com/0/0/1201782175994164


ajohn25 (approved):

ben-pr-p (approved):
I would not check in the generate types since `schema-dump.sql` is already checked in, but it depends how noisy you want your diffs versus how complex you want the build process.

I would also consider https://github.com/adelsz/pgtyped as an alternative - it generates types for your actual queries, and not just the tables that your queries might return. This means it works for joins, updates, parameter types, etc. 

It's a much stricter level of checking that makes it basically impossible to make a mistake, but it also requires a codegen step tied to changes in SQL queries instead of a codegen step tied to changes to the database.

PR #1070, opened 2022-02-10
chore(webpack): change webpack config import to TS, and use ts-node

hiemanshu:
## Description

Change webpack build config to use TS, and clean up config. Also use ts-node, with a compile to CommonJS to use TS code with node properly.

## Motivation and Context

I ran into multiple issues with getting the dev env up and running with the webpack config. 

Examples of errors: 

```
[webpack-cli] Failed to load '/Users/hiemanshu/work/rewired/Spoke/webpack.config.ts' config
[webpack-cli] /Users/hiemanshu/work/rewired/Spoke/webpack.config.ts:2
import ReactRefreshWebpackPlugin from "@pmmmwh/react-refresh-webpack-plugin";
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:979:16)
    at Module._compile (internal/modules/cjs/loader.js:1027:27)
    at Module.m._compile (/Users/hiemanshu/work/rewired/Spoke/node_modules/ts-node/src/index.ts:1056:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/hiemanshu/work/rewired/Spoke/node_modules/ts-node/src/index.ts:1059:12)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at WebpackCLI.tryRequireThenImport (/Users/hiemanshu/work/rewired/Spoke/node_modules/webpack-cli/lib/webpack-cli.js:244:16)
```

```
7:38:42 PM webpack.1        |  [webpack-cli] Failed to load '/Users/hiemanshu/work/rewired/Spoke/we
7:38:42 PM webpack.1        |  >  bpack.config.ts' config
7:38:42 PM webpack.1        |  [webpack-cli] webpack.config.ts(8,24): error TS7016: Could not find 
7:38:42 PM webpack.1        |  >  a declaration file for module './webpack/build-config'. '/Users/hiem
7:38:42 PM webpack.1        |  >  anshu/work/rewired/Spoke/webpack/build-config.js' implicitly has an 
7:38:42 PM webpack.1        |  >  'any' type.
7:38:42 PM webpack.1        |  error Command failed with exit code 2.
7:38:42 PM webpack.1        |  info Visit https://yarnpkg.com/en/docs/cli/run for documentation abo
7:38:42 PM webpack.1        |  >  ut this command.
```

## How Has This Been Tested?

Webpack now runs without issues, and clicking around in the UI hasn't caused any issues.

## Types of changes

## Documentation Changes

No changes required in documentation

## Checklist:


package.json
```diff
@@ -203,7 +203,7 @@
     "@types/lodash": "^4.14.168",
     "@types/luxon": "^1.25.2",
     "@types/material-ui": "^0.21.7",
-    "@types/node": "^14.14.30",
+    "@types/node": "^17.0.17",
```
bchrobot, line 206: Why the upgrade here? I don't believe we're using any Node 17-specific APIs

package.json
```diff
@@ -258,6 +260,7 @@
     "standard-version": "^7.1.0",
     "supertest": "^6.2.2",
     "terser-webpack-plugin": "^5.3.0",
+    "ts-node": "^10.5.0",
```
bchrobot, line 263: I'm unclear where the switch to using ts-node is actually happening. It looks like `webpack.config.ts` still uses the babel loader for .ts(x) files.

bchrobot (commented):
The dependency additions changed way more dependency resolutions in yarn.lock than I would have expected. Are there additional changes to package.json that didn't get checked in? Or is it all from ts-node?
These changes include eslint and its plugins. Some of the suggested changes are good and would bring us closer in line to React community standards but should be done in a separate PR.

package.json
hiemanshu: The nvm-version file is pinned at a different version than the types here too. I think upgrading both to newer versions of node 14 would be ideal, if the move to the 16 isn't gonna happen any time soon.

hiemanshu:
> I played around with it a little to try to get webpack to use babel-node, already included in the project and configured for TS, but it seems to really want ts-node consistent with your changes.

Yeah, I did a bit of testing around it, and this was the cleanest, most consistent way I could get it to work. `ts-node` also was the recommended way in the webpack config, and moving the webpack config import to TS, to allow imports to work without type resolution helps bring it line with the config being consistent. 

> I tried to resolve just the eslint issues in a separate PR scoped just to that but ended up running into the syntax error you had and ended up duplicating some of your changes to get things to run. Up to you whether you want to pull in the eslint changes or adopt https://github.com/politics-rewired/Spoke/pull/1072 instead.

I'd recommend pulling in the minimum required changes here to get things to work, and then pulling in the larger changes in the second PR, allowing this PR to have all the changes required to get this working.

> Yarn dependency resolution with our current mix of very old packages and very new packages feels like it may become a larger issue 😔

Indeed. I think it would be worthwhile to spend some time on this in the future, especially with making sure we have the correct node versions being run consistently across different systems, mixing up multiple versions and falling into local system issues is a common JS tooling issue.

PR #1078, opened 2022-02-15
perf: limit contact field selection

bchrobot:
## Description

Limit the fields fetched from the database to only those requested by the GraphQL query.

## Motivation and Context

[`getTexterAssignmentContactIds`](https://github.com/politics-rewired/Spoke/blob/a0e8faf3e4f45342d7e3d02e354146b72252d354/src/containers/TexterTodo.jsx#L178-L189) only requires contacts' IDs but must transfer the entire `campaign_contact` row over the wire from the database. This is slow and ties up a database connection during data transfer.

This change looks at the GraphQL query's selection set and, if only "simple" fields are requested, limits the database query's `SELECT`.

Interestingly, it seems like `getTexterAssignmentContactIds` is the only place this resolver is ever actually invoked. Using regex `/^[ ]*contacts[\( ].*\{/` we find two results. The `AdminReplySender` result is  campaign.contacts, not assignment.contacts:

<table><tr><td>

Screen Shot 2022-02-14 at 9 24 16 PM

</td><td>

Screen Shot 2022-02-14 at 9 25 27 PM

</td></tr></table>

## How Has This Been Tested?

This has been tested locally.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/api/assignment.js
+++ b/src/server/api/assignment.js
@@ -1,3 +1,4 @@
+import humps from "humps";
 import _ from "lodash";
 import request from "superagent";
 
@@ -1232,7 +1233,8 @@ export const resolvers = {
       if (selections.complex.length === 0) {
         const simpleFields = selections.simple
           .map(({ name: { value } }) => value)
-          .filter((name) => name !== "__typename");
+          .filter((name) => name !== "__typename")
+          .map((gqlKey) => humps.decamelize(gqlKey, { separator: "_" }));
         contactsQuery.select(simpleFields);
       }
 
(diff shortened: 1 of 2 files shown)


src/server/api/assignment.js
```diff
@@ -1215,7 +1216,27 @@ export const resolvers = {
...
+      // Limit fields read from DB if possible
+      const fieldNode = info.fieldNodes.find(
+        ({ name: { value } }) => value === "contacts"
+      );
+      const selections = selectionsByType(fieldNode);
+      if (selections.complex.length === 0) {
+        const simpleFields = selections.simple
+          .map(({ name: { value } }) => value)
+          .filter((name) => name !== "__typename");
+        contactsQuery.select(simpleFields);
```
ben-pr-p, line 1236: Do we not need to go from camel case to snake case for fields that aren't `id`?

ben-pr-p (approved):
Looks good pending camel case check

src/server/api/assignment.js
bchrobot: That we do! Good catch

PR #1081, opened 2022-02-16
chore: expand authz context for all user roles

ajohn25:
## Description
Adds checking permissions of all current user roles to [`AuthzProvider.tsx`](https://github.com/politics-rewired/Spoke/blob/main/src/components/AuthzProvider.tsx)

## Motivation and Context
Components that reference [`AuthzProvider.tsx`](https://github.com/politics-rewired/Spoke/blob/main/src/components/AuthzProvider.tsx) to check permissions currently can only access permissions for the Admin user role. To check other permissions, a separate GraphQL query is needed.

## How Has This Been Tested?
This has been tested locally.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] My change requires a change to the documentation.
- [ ] I have included updates for the documentation accordingly.

PR #1082, opened 2022-02-17
fix: fix authzprovider hook

ajohn25:
## Description
Fixes page render when using the new hook added in [`AuthzProvider.tsx`](https://github.com/politics-rewired/Spoke/blob/main/src/components/AuthzProvider.tsx)

## Motivation and Context
This restores previous functionality. #1081 introduced an error on render due to the use of a hook with an early return.

## How Has This Been Tested?
This has been tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/AuthzProvider.tsx
+++ b/src/components/AuthzProvider.tsx
@@ -62,6 +62,7 @@ export const AuthzProvider: React.FC<{ organizationId: string }> = (props) => {
     [roles]
   );
 
+  if (loading) return null;
   return (
     <AuthzContext.Provider value={value}>
       {props.children}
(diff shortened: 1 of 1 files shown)


ajohn25:
Error that pops up is "Rendered more hooks than during the previous render." Possibly because if [returned early](https://github.com/politics-rewired/Spoke/pull/1082/files#diff-620c7a4412e3899fc7a76f244dc4df3727e64f529b7f3356b31ed465961199b8L52) the hook is not used at all. Does that sound right? (Tested by navigating to admin/1/campaigns on main with and without the line deletions on this PR)

PR #1097, opened 2022-02-26
build: update to node v16 LTS

bchrobot:
## Description

Update to Node v16 LTS.

## Motivation and Context

We should be using LTS. The v14 requirement in package.json is also too strict to allow v16 in local development.

## How Has This Been Tested?

This has been tested locally.

N/A

## Types of changes

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "engines": {
     "npm": "please-use-yarn",
     "yarn": ">= 1.19.1",
-    "node": ">= 14.16.0"
+    "node": ">= 16.14.0"
   },
   "scripts": {
     "preinstall": "node -e \"if(process.env.npm_execpath.indexOf('yarn') === -1) throw new Error('spoke must be installed with Yarn: https://yarnpkg.com/')\"",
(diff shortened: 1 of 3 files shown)


ajohn25 (approved):

package.json
```diff
@@ -6,7 +6,7 @@
   "engines": {
     "npm": "please-use-yarn",
     "yarn": ">= 1.19.1",
-    "node": "^14.16.0"
+    "node": ">= 14.16.0"
```
hiemanshu, line 9: Suggestion: Change this to 16.14.0

Required: Also change `@types/node` to 16.x as well, so we get the types from Node 16

hiemanshu (changes requested):

package.json
bchrobot: Yeah, that sounds good!

I originally left them both at 14 to maintain compatibility with v14 but there's no reason to maintain that compatibility. Spoke isn't a library nor is our fork being widely run in production environments we don't control.

PR #1111, opened 2022-03-03
feat: allow navigation to previous and next campaign through campaign page

### LINKED ISSUE #1087: Allow navigation between higher and lower campaign IDs from campaign builder page
Navigation to other active campaign pages requires you to go to the Campaigns page. Would like to quickly navigate between campaigns.

Campaign builder should include a previous and next button to allow navigation to nearby campaign IDs.

hiemanshu:
## Description

Grab the next and previous campaign IDs, and navigate to them using button at the top of the
campaign page

## Motivation and Context

Fixes #1087

## How Has This Been Tested?

Tested locally, going through running, archived, and first and final campaigns 

Screenshot 2022-03-04 at 12 17 15 AM
Screenshot 2022-03-04 at 12 17 29 AM
Screenshot 2022-03-04 at 12 17 36 AM

## Documentation Changes

NA

## Checklist:


src/server/api/schema.js
```diff
@@ -3482,6 +3482,29 @@ const rootResolvers = {
...
+      _root,
+      { campaignId, organizationId },
+      { user }
+    ) => {
+      await accessRequired(user, organizationId, "SUPERVOLUNTEER");
+
+      const {
+        rows: [{ prevcampaign }]
+      } = await r.knex.raw(
+        `SELECT max(id) as prevCampaign FROM campaign where id < ${campaignId} AND is_started = TRUE AND organization_id = ${organizationId} AND is_archived = FALSE;`
```
ajohn25, line 3496: Since these buttons will be visible on archived campaigns, it may make sense to check is_archived for the current campaign and select prevCampaign where is_archived matches. It may be confusing if the user clicks the button on an archived campaign page, and we jump to an unarchived campaign. Defer to you if there is a better solution here though!

src/server/api/schema.js
```diff
@@ -3482,6 +3482,29 @@ const rootResolvers = {
...
+      );
+      const {
+        rows: [{ nextcampaign }]
+      } = await r.knex.raw(
+        `SELECT min(id) as nextCampaign FROM campaign where id > ${campaignId} AND is_started = TRUE AND organization_id = ${organizationId} AND is_archived = FALSE;`
```
ajohn25, line 3501: See comment above.

src/containers/AdminCampaignStats/index.jsx
```diff
@@ -108,6 +122,22 @@ class AdminCampaignStats extends React.Component {
       disableVanSyncButton: true
     });
 
+  prevCampaignClicked = () => {
```
ajohn25, line 125: Sorry, our request on #1087 should've been more specific, as we actually find it most useful to have these buttons available on the [`AdminCampaignEdit`](https://github.com/politics-rewired/Spoke/blob/main/src/containers/AdminCampaignEdit/index.jsx) page. I think it's fine to keep on the Campaign Stats page as well, as other users may find it useful here.

ajohn25 (changes requested):
UI looks good! A few notes:

src/server/api/schema.js
bchrobot, line 3501: Required: convention for SQL is to use snake case
Required: bindings should be used, especially for any user-provided values

```suggestion
      const {
        rows: [{ next_campaign }]
      } = await r.knex.raw(
        `SELECT min(id) as next_campaign FROM campaign where id > ? AND is_started = TRUE AND organization_id = ? AND is_archived = FALSE;`,
        [campaignId, organizationId]
```

src/server/api/schema.js
bchrobot, line 3496: Required: convention for SQL is to use snake case
Required: bindings should be used, especially for any user-provided values

```suggestion
        rows: [{ prev_campaign }]
      } = await r.knex.raw(
        `SELECT max(id) as prevCampaign FROM campaign where id < ? AND is_started = TRUE AND organization_id = ? AND is_archived = FALSE;`,
        [campaignId, organizationId]
```

src/api/schema.ts
```diff
@@ -225,6 +225,7 @@ const rootSchema = `
     externalLists(organizationId: String!, systemId: String!, after: Cursor, first: Int): ExternalListPage!
     notices(organizationId: String): NoticePage!
     campaignGroups(organizationId: String! after: Cursor, first: Int): CampaignGroupPage!
+    campaignNavigation(campaignId: String!, organizationId: String!): CampaignNavigation!
```
bchrobot, line 228: This can just take a campaignId and then look up the organizationId from the campaign record.

src/containers/AdminCampaignStats/index.jsx
bchrobot: The queries and button components could be pulled up into a <CampaignNavigationButtons /> component that takes campaignId as a prop.

src/server/api/schema.js
bchrobot: The buttons will also show for campaigns that have not been started yet once they're present on the campaign builder page. I think just removing the where clauses is good for now.

In the future, we could make it something like:

```
Navigating: [Started/Unstarted Dropdown] [Archived/Unarchived Dropdown]
[Previous Campaign] [Next Campaign]
```

Where the initial dropdown values are set based on the current campaign.

src/api/campaign.ts
```diff
@@ -253,6 +258,11 @@ export const schema = `
     pageInfo: PageInfo
   }
 
+  type CampaignNavigation {
+    nextCampaignId: Int
+    prevCampaignId: Int
```
bchrobot, line 263: Required: client-side IDs should always be `String`s (ref #1085)

```suggestion
    nextCampaignId: String
    prevCampaignId: String
```

bchrobot (changes requested):

src/components/CampaignNavigation.tsx
```diff
@@ -0,0 +1,77 @@
...
+        }
+      }
+    `,
+    options: (ownProps) => ({
+      variables: {
+        campaignId: ownProps.campaignId
+      }
+    })
+  }
+};
```
bchrobot, line 75: Suggestion: move query to .graphql file in [`libs/spoke-codegen/src/graphql`](https://github.com/politics-rewired/Spoke/tree/main/libs/spoke-codegen/src/graphql) to get auto-generated `useCampaignNavigationQuery()` hook instead of `loadData()` HOC.

hiemanshu:
@bchrobot I made the suggested change, and it added a little more cruft, so want to make sure you review it again before merge. Or suggest if there's a better way.

PR #1116, opened 2022-03-09
feat: allowing suspending users

### LINKED ISSUE #899: Support disabling/blocking user account

There is currently no way to remove, disable, or block a problematic user account. A user can be placed on Do Not Assign mode to prevent them requesting texting assignments but this is not an ideal solution.

There should be a way to lock a user account. This could happen instance wide (requires #783) and/or at the organization level.

N/A

**Additional context**

Move On's fork has implemented this as a "Suspended" user role (see MoveOnOrg/Spoke#1633).

hiemanshu:
Allow users to be marked as suspended, not giving them access to the organization anymore

## Description

Marking a user as suspended does the following:

1. Not allowed to be assigned texts to from the campaign edit screen
2. Cannot see any texts on the todo page
3. Cannot be assigned any texts in the message review page
4. Cannot login (will get an access denied error)

## Motivation and Context

Fixes #899 899

## How Has This Been Tested?

Tested locally for the above scenarios

## Documentation Changes

Probably needs some changes talking about suspended users, will ask Ben for help updating this

## Checklist:

### DIFF
--- a/src/containers/AdminCampaignEdit/sections/CampaignTextersForm/components/TexterAssignmentRow.tsx
+++ b/src/containers/AdminCampaignEdit/sections/CampaignTextersForm/components/TexterAssignmentRow.tsx
@@ -4,6 +4,7 @@ import DeleteIcon from "@material-ui/icons/Delete";
 import { css } from "aphrodite";
 import React from "react";
 
+import { UserRoleType } from "../../../../../api/organization-membership";
 import Slider from "../../../../../components/Slider";
 import { dataTest } from "../../../../../lib/attributes";
 import theme from "../../../../../styles/theme";
@@ -30,6 +31,7 @@ const TexterAssignmentRow: React.FC<Props> = (props) => {
 
   const {
     displayName,
+    roles,
     assignment: { contactsCount, needsMessageCount }
   } = texter;
   const messagedCount = contactsCount - needsMessageCount;
@@ -66,7 +68,9 @@ const TexterAssignmentRow: React.FC<Props> = (props) => {
       </div>
       <div className={css(rowStyles.assignedCount)}>{messagedCount}</div>
       <div {...dataTest("texterName")} className={css(rowStyles.nameColumn)}>
-        {displayName}
+        {roles?.includes(UserRoleType.SUSPENDED)
+          ? `${displayName} (SUSPENDED)`
+          : displayName}
       </div>
       <div className={css(rowStyles.input)}>
         <TextField
--- a/src/containers/AdminPeople/AdminPeople.tsx
+++ b/src/containers/AdminPeople/AdminPeople.tsx
@@ -104,6 +104,13 @@ export interface AdminPeopleMutations {
       purgeOrganizationUsers: number;
     }>
   >;
+  unassignTexts: (
+    membershipId: string
+  ) => Promise<
+    ApolloQueryResult<{
+      unassignTextsFromUser: string;
+    }>
+  >;
 }
 
 interface AdminPeopleRouteParams {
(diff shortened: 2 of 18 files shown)


src/server/api/user.js
```diff
@@ -15,12 +16,15 @@ export function buildUserOrganizationQuery(
   campaignId,
   offset
 ) {
-  const roleFilter = role ? { role } : {};
+  if (role !== UserRoleType.SUSPENDED) {
+    queryParam.whereNot({ role: UserRoleType.SUSPENDED });
+  } else if (role) {
+    queryParam.where({ role });
+  }
```
bchrobot, line 23: Question: what is the desired behavior here?

It looks like this allows either a) excluding suspended users or b) including only suspended users, but does not support filtering by other role types.

src/containers/AdminPeople/AdminPeople.tsx
```diff
@@ -199,6 +214,18 @@ class AdminPeople extends React.Component<
     this.handleCloseSuperadminDialog();
   };
 
+  handleConfirmUnassignTexts = async (unassignTexts: boolean) => {
+    const { suspendedUserMembershipId } = this.state.confirmUnassignTexts;
+    if (unassignTexts) {
+      await this.props.mutations.unassignTexts(suspendedUserMembershipId);
```
bchrobot, line 220: Question: do we check for role on the front end before allowing this action? The backend requires Owner role to perform the unassign action.

bchrobot (changes requested):
I think we also want to block suspended users from sending texts and requesting texts as well. Any other actions that should be explicitly blocked for suspended users @politics-rewired/organizing ?

src/containers/AdminCampaignEdit/sections/CampaignTextersForm/components/TexterAssignmentRow.tsx
```diff
@@ -66,7 +68,9 @@ const TexterAssignmentRow: React.FC<Props> = (props) => {
       </div>
       <div className={css(rowStyles.assignedCount)}>{messagedCount}</div>
       <div {...dataTest("texterName")} className={css(rowStyles.nameColumn)}>
-        {displayName}
+        {roles?.includes(UserRoleType.SUSPENDED)
```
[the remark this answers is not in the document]
hiemanshu: The safe chaining operator allows us to not have to worry about something being `undefined`. 

```
y = undefined
  undefined
if (y?.includes("a")) { console.log("True") } else { console.log("False") }
  False
```

ninabaldwin:
> I think we also want to block suspended users from sending texts and requesting texts as well. Any other actions that should be explicitly blocked for suspended users @politics-rewired/organizing ?

I don't have anything else that I can think of.

src/server/api/user.js
hiemanshu: If no role type is specified, get all users that are not suspended.

If role type is specified, get all users with that role. I could probably make this simpler!

src/containers/AdminPeople/AdminPeople.tsx
hiemanshu: This happens on the page where you can change user permissions, which is only access to admins and above. Should we do an additional check for this? I didn't see other permissions (like superadmin, etc) doing a check either.

src/containers/AdminCampaignEdit/sections/CampaignTextersForm/components/TexterAssignmentRow.tsx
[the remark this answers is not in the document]
bchrobot: well look at that, I've been overly cautious with this for a while now

bchrobot (approved):
Looks good!

src/containers/AdminPeople/AdminPeople.tsx
bchrobot: We can punt this to a larger role audit. Aashish noticed a few other inconsistencies elsewhere in the codebase that need to be addressed.

ajohn25 (approved):

PR #1119, opened 2022-03-11
fix: ensure non-null value for campaign isApproved

bchrobot:
## Description

This ensures a non-null value for the `isApproved` GraphQL property added in #1050.

## Motivation and Context

The migration in #1050 added a default of `false` for the `is_approved` column but did not set `not null` so there may still be null values in production.

## How Has This Been Tested?

TBD

N/A

## Documentation Changes

N/A

## Checklist:


hiemanshu:
Question: Is there a reason to not go with something simpler like `campaign.is_approved || false`

PR #1121, opened 2022-03-18
fix: handle null campaign groups for supervolunteer role

bchrobot:
## Description

This fixes the campaign list page for the Supervolunteer role.

## Motivation and Context

The `campaignGroups` resolver requires at least Admin role. This means it returns `null` for Supervolunteer. The field was not previously nullable, however, so this resulted in an error and broken page.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/api/campaign.ts
+++ b/src/api/campaign.ts
@@ -113,7 +113,7 @@ export interface Campaign {
   hasUnsentInitialMessages?: boolean | null;
   hasUnhandledMessages?: boolean | null;
   teams: Team[];
-  campaignGroups: CampaignGroupPage | null;
+  campaignGroups?: CampaignGroupPage | null;
   externalSystem?: ExternalSystem | null;
   creator?: User | null;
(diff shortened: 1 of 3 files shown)


src/api/campaign.ts
```diff
@@ -113,7 +113,7 @@ export interface Campaign {
   hasUnsentInitialMessages?: boolean | null;
   hasUnhandledMessages?: boolean | null;
   teams: Team[];
-  campaignGroups: CampaignGroupPage;
+  campaignGroups: CampaignGroupPage | null;
```
hiemanshu, line 116: Nitpick: All optional variables end with a `?`

hiemanshu (approved):
LGTM!

src/api/campaign.ts
bchrobot: That is the convention! I will change

Issue #409, opened 2019-12-23
Texter notifications should be configurable


bchrobot:
Texters should have the option of receiving digests of replies, or some other method of batched notifications after a period of Texter inactivity.

There would be four levels of notification frequency:

- daily digest
- periodic digest (cluster in ~4 hour buckets)
- one email for every notification
- off
​
At first this would be one setting for all types of notifications. Eventually we might want to allow setting preference for each type of notification.

hiemanshu:
Some thoughts / ramblings:

1. Add a new column in `user_organization` where value is picked from an enum
2. Create a new table to store digest messages
3. Allow changing in account page, with default as an email for each notification
4. When a notification is created, if it should be sent, send it right away, otherwise add to digest table
5. Worker processes users every 4 hours for periodic digest, deletes messages from digest table when done
6. Worker processed every 24 hours for daily digest, deletes messages from digest table when done

@bchrobot Just some thoughts on these, do you have anything to add here? Or any feedback?

hiemanshu:
> My only thought is whether everything should go to the digest table and notifications should be sent 1 minute after they're created to allow time for an admin to "undo" the notification-triggering action (e.g. unstart a campaign, delete assignment). We don't really have these undo mechanisms in place, however, so this can also be saved for future work.

This sounds like a good plan! Just having everything go through the same mechanisms, just handled by differently timed workers is a good idea! This can also allow frontend notifications through the browser, other channels, and even undo in the future. 

> I think there could just be a sent_at timestamptz column in the notifications table that starts null and is set once the digest email for the notification is sent. Future work could add a notifications menu dropdown to access the notifications in-app.

I was trying to think of possibly ending up with a super large and unwieldy type notifications table. I guess setting up periodic archiving might be better, and allow a lot more flexibility, something we can deal with when it becomes larger.

hiemanshu:
Going to split this into 2 features:

1. Create a schema, and table for notifications, and create a worker to process it
2. Create workers, and UI to allow sending digest notifications.

ninabaldwin:
@bchrobot @hiemanshu 

To answer the two questions above:
1. Let's do instance-wide
3. Default as email for each notification, but we like the email to specify that there is an option to change this to a digest

hiemanshu:
@bchrobot @politics-rewired/organizing I've made a Draft PR here: https://github.com/politics-rewired/Spoke/pull/1131

I'd like to get feedback on the implement from Ben.

[...]

PR #1126, opened 2022-03-23
feat: use worker to process notifications

hiemanshu:
This is part one of texter notification changes. Made this separate so we can talk about schema, and possible changes to support future features.

## Description

Write notifications to database and process using a worker task, part 1 of adding digest emails

## Motivation and Context

This is part one of #409 

## How Has This Been Tested?

Tested locally.

## Documentation Changes

NA

## Checklist:


src/server/tasks/send-notifications.ts
```diff
@@ -0,0 +1,32 @@
...
+        replyTo: notification.replyTo
+      });
+
+      await r
+        .knex("notification")
+        .where({ id: notification.id })
+        .update({ sent_at: new Date() });
+    } catch (e) {
+      console.log(e);
+      console.log("Failed to send email notification...");
```
bchrobot, line 27: Required: Use `winston` for structured logging:

```suggestion
      logger.error('Failed to send email notification: ', errToObj(e));
```

For reference, see `auth-passport.ts`:
https://github.com/politics-rewired/Spoke/blob/8099f8caaa80899bfb88feee87abfbb745c7f4fd/src/server/auth-passport.ts#L85-L87

src/server/tasks/send-notifications.ts
```diff
@@ -0,0 +1,32 @@
+import { Task } from "pg-compose";
+
+import { sendEmail } from "../mail";
+import { r } from "../models";
+
+const sendNotifications: Task = async (_payload, _helpers) => {
```
bchrobot, line 6: Required: Our approach for fan-out tasks like this has been to have one parent task (e.g. `queue-pending-notifications`) that creates individual jobs for each notification (e.g `send-notification`).

This provides more granular error information, better handles upstream throttling, and means that child tasks can be created by the parent using the [`addJob` parameter](https://github.com/graphile/worker#addjob) to deduplicate tasks should it end up taking more than 1 minute to process all notifications from the previous minute.

It may be easier to write the body of the parent task as SQL statement (or just use `knex.raw` for the select in the knex query builder chain):

```sql
select graphile_worker.add_job('send-notification-email', json_build_object('to', to, 'subject', subject, 'text', content, 'replyTo', reply_to))
from notification
join public.user on user.id = notification.user_id
where sent_at is null
```

bchrobot (changes requested):

bchrobot (changes requested):

bchrobot (approved):

ajohn25 (approved):

PR #1127, opened 2022-03-24
fix: fix permissions in script preview url resolver

bchrobot:
## Description

Fix permission checks in resolver for `Campaign.previewUrl`.

## Motivation and Context

Message Review uses the `loadData()` HOC which does not allow for ignoring expected GraphQL errors. We must explicitly return `null` for `previewUrl` if the permissions check fails.

The permissions check also needs to look at the organization setting to determine which role is sufficient.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/api/campaign.js
+++ b/src/server/api/campaign.js
@@ -12,6 +12,7 @@ import { getDeliverabilityStats } from "./lib/campaign";
 import { symmetricEncrypt } from "./lib/crypto";
 import { formatPage } from "./lib/pagination";
 import { sqlResolvers } from "./lib/utils";
+import { getOrgFeature } from "./organization-settings";
 import { getUsers } from "./user";
 
 export function addCampaignsFilterToQuery(queryParam, campaignsFilter) {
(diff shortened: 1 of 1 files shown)


src/server/api/campaign.js
```diff
@@ -500,11 +501,21 @@ export const resolvers = {
...
-      await accessRequired(user, organizaitonId, "ADMIN");
+    previewUrl: async (campaign, _, { user, loaders }) => {
+      const campaignId = campaign.id;
+      const organizationId = await getCampaignOrganization({ campaignId });
+      const { features } = await loaders.organization.load(organizationId);
+      const supervolAccess = getOrgFeature(
+        "scriptPreviewForSupervolunteers",
+        features
+      );
+      const requiredRole = supervolAccess ? "SUPERVOLUNTEER" : "ADMIN";
```
hiemanshu, line 512: Suggestion: User `UserRoleType`

hiemanshu (approved):

src/server/api/campaign.js
bchrobot: ooh, good call! Made the change

ajohn25 (approved):

PR #1131, opened 2022-03-30
feat(notifications): allow texters to configure notification frequency

hiemanshu:
## Description

Allow user to to choose between different notification frequencies. ALL sends all the notifications,
PERIODIC sends it at 9AM, 1PM, 5PM, and 9PM, DAILY sends it at 9AM daily

Creating this PR early to gather feedback, and sort out details.

Need the following before fully merging:

1. Possible updates to the UserEdit UI
2. Notification Template
3. Notification Digest Template

## Motivation and Context

Context: #409 

## How Has This Been Tested?

Tested Locally

WIP. Coming soon.

## Checklist:

### DIFF
--- a/src/api/user.ts
+++ b/src/api/user.ts
@@ -8,6 +8,12 @@ import type {
 import type { RelayPaginatedResponse } from "./pagination";
 import type { Team } from "./team";
 
+export enum NotificationFrequencyType {
+  ALL = "ALL",
+  PERIODIC = "PERIODIC",
+  DAILY = "DAILY"
+}
+
 export interface User {
   id: string;
   firstName: string;
@@ -25,6 +31,7 @@ export interface User {
   assignment: Assignment;
   terms: boolean;
   isSuperadmin: boolean;
+  notificationFrequency: NotificationFrequencyType;
 }
 
 export const schema = `
@@ -42,9 +49,10 @@ export const schema = `
     teams(organizationId: String!): [Team]!
     currentRequest(organizationId: String!): AssignmentRequest
     assignedCell: Phone
-    assignment(campaignId: String): Assignment,
+    assignment(campaignId: String): Assignment
     terms: Boolean
     isSuperadmin: Boolean!
+    notificationFrequency: String
   }
   
   type UsersList {
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -487,7 +487,8 @@ const rootMutations = {
           first_name: userData.firstName,
           last_name: userData.lastName,
           email: userData.email,
-          cell: userData.cell
+          cell: userData.cell,
+          notification_frequency: userData.notificationFrequency
         });
 
         memoizer.invalidate(cacheOpts.GetUser.key, { id: userId });
@@ -500,7 +501,8 @@ const rootMutations = {
           first_name: userData.firstName,
           last_name: userData.lastName,
           email: userData.email,
-          cell: userData.cell
+          cell: userData.cell,
+          notification_frequency: userData.notificationFrequency
         };
       } else {
         userData = member;
--- a/src/server/api/user.js
+++ b/src/server/api/user.js
@@ -179,6 +179,9 @@ export const resolvers = {
 
       return currentRequest || null;
     },
+    notificationFrequency: async (user, _params, _) => {
+      return user.notification_frequency;
+    },
     assignment: async (user, { campaignId }) =>
       r
         .reader("assignment")
--- a/src/server/lib/templates/digest.tsx
+++ b/src/server/lib/templates/digest.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+import ReactDOMServer from "react-dom/server";
+
+interface Notification {
+  id: number;
+  subject: string;
+  content: string;
+}
+
+interface Props {
+  notifications: Notification[];
+}
+
+const Digest: React.FC<Props> = ({ notifications }) => {
+  return (
+    <div>
+      <p>Here's your periodic digest, with all the notifications for you!</p>
+      {notifications.map((notification) => {
+        return (
+          <div key={notification.id}>
+            <p style={{ fontWeight: "bold" }}>{notification.subject}</p>
+            <br />
+            <br />
+            <p>{notification.content}</p>
+          </div>
+        );
+      })}
+    </div>
+  );
+};
+
+const getDigest = (notifications: Notification[]) => {
+  const template = <Digest notifications={notifications} />;
+  return ReactDOMServer.renderToStaticMarkup(template);
+};
+
+export default getDigest;
--- a/src/server/tasks/queue-pending-notifications.ts
+++ b/src/server/tasks/queue-pending-notifications.ts
@@ -2,10 +2,20 @@ import { Task } from "pg-compose";
 
 import { r } from "../models";
 
-const queuePendingNotifications: Task = async (_payload, helpers) => {
+const getUserIdsForDigest = (frequency: string) => {
+  return r
+    .knex("notification")
+    .whereNull("sent_at")
+    .where("user.notification_frequency", frequency)
+    .select(["user.id"])
+    .join("user", "user_id", "user.id");
+};
+
+export const queuePendingNotifications: Task = async (_payload, helpers) => {
   const notificationsToSend = await r
     .knex("notification")
     .whereNull("sent_at")
+    .where("notification_frequency", "ALL")
     .select(["notification.id", "email", "subject", "content", "reply_to"])
     .join("user", "user_id", "user.id");
 
@@ -16,4 +26,22 @@ const queuePendingNotifications: Task = async (_payload, helpers) => {
   }
 };
 
-export default queuePendingNotifications;
+export const queuePeriodicNotifications: Task = async (_payload, helpers) => {
+  const usersToNotify = await getUserIdsForDigest("PERIODIC");
+
+  for (const user of usersToNotify) {
+    await helpers.addJob("send-notification-digest", user, {
+      jobKey: `send-notification-periodic-${user.id}`
+    });
+  }
+};
+
+export const queueDailyNotifications: Task = async (_payload, helpers) => {
+  const usersToNotify = await getUserIdsForDigest("DAILY");
+
+  for (const user of usersToNotify) {
+    await helpers.addJob("send-notification-digest", user, {
+      jobKey: `send-notification-daily-${user.id}`
+    });
+  }
+};
--- a/src/server/tasks/send-notification-email.ts
+++ b/src/server/tasks/send-notification-email.ts
@@ -1,17 +1,34 @@
 import { Task } from "pg-compose";
 
 import logger from "../../logger";
+import getDigest from "../lib/templates/digest";
 import { sendEmail } from "../mail";
 import { r } from "../models";
 import { errToObj } from "../utils";
 
-const sendNotificationEmail: Task = async (payload, _helpers) => {
+const createDigest = async (userId: number) => {
+  const notifications = await r
+    .knex("notification")
+    .where({ user_id: userId })
+    .whereNull("sent_at")
+    .select(["notification.id", "email", "subject", "content", "reply_to"])
+    .join("user", "user_id", "user.id");
+
+  const html = getDigest(notifications);
+
+  return {
+    html,
+    notifications
+  };
+};
+
+export const sendNotificationEmail: Task = async (payload, _helpers) => {
   try {
-    const { email, subject, text, replyTo } = payload;
+    const { email, subject, content, replyTo } = payload;
     await sendEmail({
       to: email,
       subject,
-      text,
+      text: content,
       replyTo
     });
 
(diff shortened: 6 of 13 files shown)


src/api/user.ts
```diff
@@ -42,9 +49,10 @@ export const schema = `
     teams(organizationId: String!): [Team]!
     currentRequest(organizationId: String!): AssignmentRequest
     assignedCell: Phone
-    assignment(campaignId: String): Assignment,
+    assignment(campaignId: String): Assignment
     terms: Boolean
     isSuperadmin: Boolean!
+    notificationFrequency: String
```
bchrobot, line 55: Required: this should be a non-nullable field. You might need to tweak the migration to set that on the column as well. `not null` + `default` should backfill all existing rows.

src/api/user.ts
```diff
@@ -8,6 +8,12 @@ import type {
 import type { RelayPaginatedResponse } from "./pagination";
 import type { Team } from "./team";
 
+export enum NotificationFrequencyType {
+  ALL = "ALL",
+  PERIODIC = "PERIODIC",
+  DAILY = "DAILY"
```
bchrobot, line 14: Suggestion/required: The prevailing trend is to use upper camel case for enum keys

Values are pretty mixed between uppercase (GraphQL convention) and lowercase (Postgres enum type convention), unfortunately.

src/server/api/schema.js
```diff
@@ -487,7 +487,8 @@ const rootMutations = {
           first_name: userData.firstName,
           last_name: userData.lastName,
           email: userData.email,
-          cell: userData.cell
+          cell: userData.cell,
+          notification_frequency: userData.notificationFrequency
```
bchrobot, line 491: Required: `notificationFrequency` should either also be a required input property or we should only include in update if it's present in the payload (otherwise we could be setting it to `null`)

```suggestion
          ...(userData.notificationFrequency ? { notification_frequency: userData.notificationFrequency } : {})
```

src/server/api/user.js
```diff
@@ -179,6 +179,9 @@ export const resolvers = {
 
       return currentRequest || null;
     },
+    notificationFrequency: async (user, _params, _) => {
```
bchrobot, line 182: Suggestion: This can go in `...sqlResolvers([` above if it's just a snake case to camel case conversion

src/server/lib/templates/digest.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import ReactDOMServer from "react-dom/server";
+
+interface Notification {
+  id: number;
+  subject: string;
+  content: string;
+}
```
bchrobot, line 8: Suggestion: This looks like it could be sourced from `libs/spoke-codegen/src/pg-types.ts` (via `import from "@spoke/spoke-codegen"`).

Required: You'll just need to run `yarn codegen:pg` first (and `./update-dump.sh` whenever there are schema changes more generally).

src/server/tasks/send-notification-email.ts
```diff
@@ -24,4 +41,25 @@ const sendNotificationEmail: Task = async (payload, _helpers) => {
...
+      html
+    });
+
+    await r
+      .knex("notification")
+      .whereIn(
+        "id",
+        notifications.map((n) => n.id)
+      )
+      .update({ sent_at: new Date() });
```
bchrobot, line 61: Suggestion: I think wrapping both in a transaction and performing the update first is better.

That way it's only a `commit` between a successful `sendEmail()` and everything being updated, rather than the entire update command after `sendEmail()`.

bchrobot (commented):
Very excited to do more with `ReactDOMServer.renderToStaticMarkup` emails!

hiemanshu:
Just double checking through things, will add screenshots to these soon!

hiemanshu:
@politics-rewired/organizing here are screenshots of what this looks like! 

Screenshot 2022-04-14 at 3 00 17 PM
Screenshot 2022-04-14 at 3 00 23 PM
Screenshot 2022-04-14 at 3 00 29 PM
Screenshot 2022-04-14 at 3 00 37 PM
Screenshot 2022-04-14 at 3 00 47 PM
Screenshot 2022-04-14 at 3 01 06 PM

And here are what the emails look like! These can be spruced up real nice, but that's for another feature.

Screenshot 2022-04-14 at 3 12 03 PM
Screenshot 2022-04-14 at 3 12 18 PM
Screenshot 2022-04-14 at 3 09 18 PM

migrations/20220406135225_change_notification_to_category.js
```diff
@@ -0,0 +1,31 @@
+exports.up = function up(knex) {
+  return knex.schema.alterTable("notification", (table) => {
+    table.dropColumn("subject");
+    table.dropColumn("content");
+    table.dropColumn("reply_to");
+
+    table.integer("organization_id");
+    table.integer("campaign_id");
```
bchrobot, line 8: Question: how do we handle future notification types?

Based on a nullable `campaign_id`, it seems like you're envisioning adding more nullable columns as needed for new notification types, is that correct?

Something else to consider is adding a more dynamic payload column (instead of or in addition to). Schema would have to be enforced at the application level and values in here couldn't have foreign key constraints). Versioned notification schemas could help address the first issue. Keeping properties like `campaign_id` as nullable columns while putting other properties like `imageUrl` in JSON could address the second issue.

```ts
table.jsonb("payload");
```

bchrobot (commented):
Required: Fetching data and rendering a template should be separated.

In my mind, that could look like something with the following three pieces:

1. Versioned TS interfaces for each notification payload
    ```ts
    interface AssignmentCreatedNotificationV1 {
      type: "AssignmentCreatedNotificationV1";
      assignmentId: number;
      campaignId: number;
      campaignTitle: string;
      assignmentCount: number;
    }

    interface AssignmentUpdatedNotificationV1 {
      type: "AssignmentUpdatedNotificationV1";
      assignmentId: number;
      campaignId: number;
      campaignTitle: string;
      assignmentCount: number;
    }

    // ...

    type Notification =
      | AssignmentCreatedNotificationV1
      | AssignmentUpdatedNotificationV1;
      // ...
    ```
2. Functions to fetch data
    ```ts
    function getNotificationsForUser(userId: number, frequency: NotificationFrequencyType): Promise<Notification[]> = {}
    ```
3. Templates that consume `Notification`s:
    ```ts
    interface AssignmentCreatedDigestRowProps {
      notification: AssignmentCreatedNotification;
    } 
    const AssignmentCreatedDigestRow: React.FC<AssignmentCreatedDigestRowProps> = (props) => {
      const { campaignTitle, assignmentCount } = props.notification;
      return (
        <p>`[${campaignTitle}]: New Assignment. <b>${assignmentCount}</b> messages to send`<p>
      );
    };
    
    interface DigestEmailProps {
      notifications: Notification[];
    }
    
    const DigestEmail: React.FC<DigestEmailProps> = (props) => {
      return (
        <ul>
          {props.notifications.map((notification) => {
            switch (notification.type) {
              case "AssignmentCreatedNotification":
                return <AssignmentCreatedDigestRow key="assignment-created" notification={notification} />
            }
          })};
        </ul>
      );
    };
    ```

The versioning allows us to change notification schema in backwards compatible way (see comment in migration below).

migrations/20220406135225_change_notification_to_category.js
hiemanshu: My assumption here was that most notifications will likely be connected to a campaign / organization. i.e. An org has suspended you, or you have something in campaign, or a sweeper needs to look at something in org. We can add an extra payload column in the future for extra data, and campaign, and organization being indexable would improve performance when looking at some of those views.

src/server/tasks/queue-pending-notifications.ts
```diff
@@ -2,11 +2,28 @@ import { Task } from "pg-compose";
 
 import { r } from "../models";
 
-const queuePendingNotifications: Task = async (_payload, helpers) => {
+const getUserIdsForDigest = (frequency: string) => {
```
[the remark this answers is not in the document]
hiemanshu: We use full capital names in both the frontend and backend, so we can just use the same types for now. The convention even for psql has been to use full caps, so "PERIODIC" works for both GraphQL and PSql!

migrations/20220406135225_change_notification_to_category.js
```diff
@@ -0,0 +1,31 @@
...
+    table.string("category").notNullable().default("");
```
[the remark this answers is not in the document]
hiemanshu: Yeah, this is something I'm split on. Usually following convention/paradigms (in other languages too), type style variable names are kept so you could have multiple Models `NotificationType1`, `NotificationType2`, etc accessing the same table based on the type variable. It's why I went with category. I don't feel strongly towards either name, let me know what you prefer, and I'll just go with that!

libs/spoke-codegen/src/pg-types.ts
```diff
@@ -1,22 +1,22 @@
 export enum messaging_service_type {
-  'Twilio' = 'twilio',
-  'AssembleNumbers' = 'assemble-numbers' 
+  'AssembleNumbers' = 'assemble-numbers',
+  'Twilio' = 'twilio' 
 }
```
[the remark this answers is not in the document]
hiemanshu: Do you have anything for this yet?

hiemanshu:
For the split between in notification data fetching / rendering, how do you see the files being separated? I spent a bit of time thinking about this, and the email sender shouldn't have to worry about fetching the data, so it was passed onto the component file to fetch it's own data. Usually we fetch the data in the component (or container) and render it. 

Agree with you on the types being moved out with versioning. Just trying to think what file will live where if we split up the data fetching, and the data rendering parts.

migrations/20220406135225_change_notification_to_category.js
bchrobot: Sounds good!

libs/spoke-codegen/src/pg-types.ts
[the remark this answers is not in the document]
bchrobot: `campaign_with_groups` is indeed local drift.

I ran all migrations on a fresh DB, ran `codegen:pg`, and got a huge diff 😔

I think removing pg-types autogeneration is what has to happen (in its own PR).

src/server/api/schema.js
bchrobot: Addressed in 9588dbdf22c8cecd1a09e7fbd227218102a69c52

migrations/20220406135225_change_notification_to_category.js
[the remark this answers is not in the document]
bchrobot: I think `notification_type` that refers to a specific version (e.g. `notification_type = 'assignment_created_v1'` is the most future proof in the event we do add a JSON(B) column type. We'll need to know what version of the schema the JSON matches.

libs/spoke-codegen/src/pg-types.ts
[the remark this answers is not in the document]
hiemanshu: Should I remove the auto type in this PR if it's going to be removed in the future?

hiemanshu:
> In this context, SSR React just happens to be the templating engine we're using to render the emails. I think it's helpful to think about it not as a component, but strictly as a template, and not have it fetch its own data.

Fair enough. 

> When you say email sender, do you mean the sendEmail function or the tasks? I think the tasks in src/server/tasks/send-notification-email.ts are the right place to fetch data and pass to the "template engine" for rendering. Or, if not in the task body itself, an wrapper function that takes the notification ID or user ID and a PoolClient and does the fetching and rendering there.

In the tasks, yes. The file is already starting to get large, and I think when more types of notifications get added it will get larger, which is why I wanted the data fetching happening around the template. I think the best way to keep most of it atomic, including versioning support, would be to have a helper function in it's own file, that fetches the data, and populates the templates with the content and returns the content. So we could switch the helper function depending on versioning, and all the large fetching / wrangling happens in it's own file, so we don't end up with super large files that encompass everything.

hiemanshu:
@bchrobot I've made the large refactor to move data wrangling to it's own module. This is ready for a re-review!

bchrobot (approved):
Looks good!

src/components/SaveNotificationSettingsAlert.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const SaveNotificationSettingsAlert: React.FC<SaveNotificationSettingsAlertProps> = (
+  props
+) => {
+  const getSnackbarText = () => {
+    switch (props.notificationFrequency) {
+      case NotificationFrequencyType.All:
+        return (
+          <p>
+            Emails will be sent for all notification, including assignment
```
ajohn25, line 24: Nitpick:
```suggestion
            Emails will be sent for all notifications, including assignment
```

ajohn25 (changes requested):
Some copy nitpicks but otherwise looks great!

src/components/SaveNotificationSettingsAlert.tsx
```diff
@@ -0,0 +1,80 @@
...
+        return (
+          <p>
+            Email notifications are sent at 9AM, 1PM, 4PM, and 8PM in the time
+            zone of your Spoke workspace.
+          </p>
+        );
+      case NotificationFrequencyType.None:
+        return (
+          <p>
+            You will not receive notifications about changes to your assignment,
```
ajohn25, line 45: Nitpick:

```suggestion
            You will not receive notifications about changes to your assignments,
```

ajohn25 (approved):

PR #1140, opened 2022-04-09
build: use node compatible with v16

ajohn25:
## Description
Require Node compatible with v16.

## Motivation and Context
#1097 updates package.json to allow Node versions greater than v16 LTS. However, we are still using packages that rely on deprecated methods, which produces errors if using current release (v17) in local development. Additionally, we would expect similar errors if using v18 when it is released (soon) and later becomes LTS.

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "engines": {
     "npm": "please-use-yarn",
     "yarn": ">= 1.19.1",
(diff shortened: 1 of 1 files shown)


bchrobot (approved):
Good catch!

PR #1157, opened 2022-04-27
feat: support user session management

bchrobot:
## Description

This adds support managing user sessions and adds an instance-level suspended flag.

It also moves user actions into a dropdown menu.

## Motivation and Context

Continues work of #899.

## How Has This Been Tested?

This has been tested locally and unit tests have been added.

<table>
<tr><td>

Fullscreen_4_27_22__9_53_AM

</td></tr><tr><td>

Politics_Rewired_-_People

</td></tr>
</table>

## Documentation Changes

These changes should be documented, but are not.

## Checklist:

---
  - https://app.asana.com/0/0/1202182602709205


src/server/api/schema.js
```diff
@@ -556,6 +557,38 @@ const rootMutations = {
...
+    setUserSuspended: async (_root, { userId, isSuspended }, { user, db }) => {
+      await superAdminRequired(user);
+
+      await db.primary.transaction(async (trx) => {
+        await trx.raw(`update public.user set is_suspended = ? where id = ?`, [
+          isSuspended,
+          userId
+        ]);
+
+        await trx.raw(`delete from user_session where user_id = ?`, [userId]);
```
ben-pr-p, line 569: Is it ok to do this even if what we're doing here is "unsuspending"? I think so, but just thinking aloud

ben-pr-p (approved):
Code looks good –

Do we need extra handling around the transition?

Looking at the code, it looks like because we are removing the `cookieSession` middleware, we will just start treating users who are logged in as not logged in, and standard application behavior will force them to log in.

Were you able to test that locally, and check we don't get into any stuck state?

src/server/api/schema.js
bchrobot: The delete shouldn't actually do anything here.

All of the user's sessions will have been deleted when they were suspended and they won't have been able to log in to create new ones between when they were suspended and un-suspended.

ajohn25 (approved):

migrations/20220426234444_support-suspending-users.js
```diff
@@ -0,0 +1,23 @@
+// Ref: https://github.com/voxpelli/node-connect-pg-simple/blob/main/table.sql
+
+exports.up = function up(knex) {
+  return knex.schema.raw(`
+    alter table public.user add column is_suspended boolean not null default false;
+
+    create table user_session (
+        sid text primary key
+      , sess json not null
```
hiemanshu, line 9: Nitpick: SQL lines end with a comma, and not start with one, it's also the style we've been following so far

hiemanshu (approved):
LGTM! Just a small nitpick overall.

migrations/20220426234444_support-suspending-users.js
bchrobot: Good catch!

PR #1187, opened 2022-05-12
fix: throw error and retry on empty email

bchrobot:
## Description

This adds retry behavior to notification email sending.

## Motivation and Context

It may be the case that generating email content is either

1. not being awaited correctly (see [SO post]), or
2. returning empty content some of the time

We should throw an error and retry in both these cases.

We should also throw an error for unrecognized notification template types rather than returning a an empty `<div />` as content.

[SO post]: https://stackoverflow.com/a/54484820

## How Has This Been Tested?

I was able to reproduce the empty message error locally only once but was not able to capture enough detail to debug why the content is empty.

N/A

## Documentation Changes

N/A

## Checklist:

---
  - https://app.asana.com/0/0/1202266689193943

### DIFF
--- a/src/server/lib/templates/notification.tsx
+++ b/src/server/lib/templates/notification.tsx
@@ -114,7 +114,7 @@ const getNotificationContent = (
       subject = `[${organization.name}] New Reply: ${campaign.title}`;
       break;
     default:
-      template = <div />;
+      throw new Error(`Unrecognized notification type ${notificationType}`);
   }
 
   const content = ReactDOMServer.renderToStaticMarkup(template);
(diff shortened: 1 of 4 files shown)


ajohn25 (approved):
Heads up that "(see [SO post](https://github.com/politics-rewired/Spoke/pull/1))" in the Motivation and Context here just links to the first Spoke PR 🙂 Retry seems to make sense either way though.

hiemanshu (approved):

hiemanshu:
Do you think the empty emails were because there could have been some notifications that weren't sent, but didn't have the correct notification category applied to it, since the migration would remove a bunch of the old columns and add new ones without the new ones having data.

PR #1190, opened 2022-05-13
fix: fix csv header validation

### LINKED ISSUE #1133: Uploading a contact list with "firsstName" column and upload is stuck
**Describe the bug**
The client tried to upload a contact list and added an extra S in the firstName column, making it "firsstName." The problem is that the upload appeared to be stuck and there was no error indication.

**To Reproduce**

1.  Upload a csv with one of the columns reading firsstName 
1.  File is stuck, received no error message

**Expected behavior**

User receives an error and is able to correct the file.

bchrobot:
## Description

This PR does two things:

1. Trims whitespace around column names
2. Fixes error display in the client

## Motivation and Context

Closes #1133.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/containers/AdminCampaignEdit/sections/CampaignContactsForm/index.tsx
+++ b/src/containers/AdminCampaignEdit/sections/CampaignContactsForm/index.tsx
@@ -140,7 +140,7 @@ class CampaignContactsForm extends React.Component<
         excludeCampaignIds: selectedCampaignIds
       };
       const response = await this.props.mutations.editCampaign(campaignInput);
-      if (response.errors) throw response.errors;
+      if (response.errors) throw response.errors[0];
       // TODO: this.props.onComplete();
     } catch (err) {
       this.props.onError(err.message);
--- a/src/server/api/lib/edit-campaign.js
+++ b/src/server/api/lib/edit-campaign.js
@@ -46,12 +46,13 @@ export const processContactsFile = async (file) => {
     Papa.parse(stream, {
       header: true,
       transformHeader: (header) => {
+        const trimmedHeader = header.trim();
         for (const [field, aliases] of Object.entries(fieldAliases)) {
-          if (aliases.includes(header)) {
+          if (aliases.includes(trimmedHeader)) {
             return field;
           }
         }
-        return header;
+        return trimmedHeader;
       },
       step: ({ data, meta, errors: _errors }, parser) => {
         // Exit early on bad header


hiemanshu (approved):

src/containers/AdminCampaignEdit/sections/CampaignContactsForm/index.tsx
```diff
@@ -140,7 +140,7 @@ class CampaignContactsForm extends React.Component<
         excludeCampaignIds: selectedCampaignIds
       };
       const response = await this.props.mutations.editCampaign(campaignInput);
-      if (response.errors) throw response.errors;
+      if (response.errors) throw response.errors[0];
```
ajohn25, line 143: Question: Were you able to figure out if this was needed because of changes in #1027 or somewhere else?

ajohn25 (approved):

src/containers/AdminCampaignEdit/sections/CampaignContactsForm/index.tsx
bchrobot: I think it's actually more likely that it was broken in #644 but I haven't stepped through to confirm that.

bchrobot: [v2 ApolloQueryResult](https://www.apollographql.com/docs/react/v2/api/apollo-client#types) has both `error` and `errors` properties. v3 exports ([1](https://github.com/apollographql/apollo-client/blob/dc0c7d1e3a9e43452737be17c3b75c90d1d7130c/src/core/QueryManager.ts#L178), [2](https://github.com/apollographql/apollo-client/blob/dc0c7d1e3a9e43452737be17c3b75c90d1d7130c/src/link/core/types.ts#L23-L31)) `graphql`'s [`ExecutionResult`](https://github.com/graphql/graphql-js/blob/52231d1b61911c2de05b06018e629afb412fd61a/src/execution/execute.ts#L127-L134) which only has `errors`.

They redid the documentation site recently with better v2 support but for a while it was tricky to find good answers about how to do what we wanted, especially with the `loadData()` HOC, with v2. I think we just got sloppy/confused by v3 documentation and started throwing `errors` as standard practice. `loadData()` handles that okay but in other situations, like here, it didn't work but also didn't noisily break anything so we didn't notice.

PR #1202, opened 2022-05-15
fix: cleanup autosending counts

ajohn25:
## Description

This PR makes 3 changes to autosending:
- Changes the "Waiting to Send" count to exclude opted out contacts (using a query)
- Prevents starting autosending on a campaign where there are no more initials
- Modifies the behavior of queue-autosend-initials to mark a campaign as complete if there are no more contacts with 0 sending attempts for the campaign

## Motivation and Context
Feedback from user testing.

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/src/api/campaign.ts
+++ b/src/api/campaign.ts
@@ -149,6 +149,7 @@ export const schema = `
     sentMessagesCount: Int
     receivedMessagesCount: Int
     optOutsCount: Int
+    needsMessageOptOutsCount: Int
     percentUnhandledReplies: Float!
     countMessagedContacts: Int!
   }
--- a/src/api/campaign.ts
+++ b/src/api/campaign.ts
@@ -152,6 +152,7 @@ export const schema = `
     needsMessageOptOutsCount: Int
     percentUnhandledReplies: Float!
     countMessagedContacts: Int!
+    countNeedsMessageContacts: Int!
   }
 
   type DeliverabilityErrorStat {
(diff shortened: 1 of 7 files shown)


src/server/api/campaign.js
```diff
@@ -245,6 +245,32 @@ export const resolvers = {
...
+                and archived = ${archived}
+                and campaign_id = ?
+            `,
+            [campaignId]
+          );
+
+          const [{ count_needs_message_contacts: result }] = rows;
+          return result;
+        },
+        cacheOpts.PercentUnhandledReplies
```
ben-pr-p, line 265: We need a new cacheOpts here - otherwise, we'll cache both this and percentUnhandledReplies with the same key, and they'll overwrite each other!

ben-pr-p (approved):
Looks good to me after the cacheOpts change!

src/api/campaign.ts
```diff
@@ -149,8 +149,10 @@ export const schema = `
     sentMessagesCount: Int
     receivedMessagesCount: Int
     optOutsCount: Int
+    needsMessageOptOutsCount: Int
```
bchrobot, line 152: Requirement: the resolver for this should always return a value

```suggestion
    needsMessageOptOutsCount: Int!
```

bchrobot (changes requested):

PR #1208, opened 2022-05-16
fix: turn off initial autoassignment for campaigns that are autosending

ajohn25:
## Description
This prevents initials on campaigns that are actively being autosent from being assigned.

## Motivation and Context
Since autosending unassigns assigned initials, it can be confusing for texters to pick up initials for a campaign that is being autosent.

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/migrations/20220516023025_exclude-autosend-campaigns-from-autoassign.js
+++ b/migrations/20220516023025_exclude-autosend-campaigns-from-autoassign.js
@@ -1,0 +1,58 @@
+exports.up = function up(knex) {
+  return knex.schema
+    .raw(
+      `
+          create or replace view assignable_campaigns as (
+            select id, title, organization_id, limit_assignment_to_teams, autosend_status
+            from campaign
+            where is_started = true
+              and is_archived = false
+              and is_autoassign_enabled = true
+              and texting_hours_end > extract(hour from (CURRENT_TIMESTAMP at time zone campaign.timezone))
+          )
+        `
+    )
+    .then(() => {
+      return knex.schema.raw(`
+          create or replace view assignable_campaigns_with_needs_message as (
+            select *
+            from assignable_campaigns
+            where exists (
+              select 1
+              from assignable_needs_message
+              where campaign_id = assignable_campaigns.id
+            )
+            and autosend_status <> 'sending'
+          );
+        `);
+    });
+};
+
+exports.down = function down(knex) {
+  return knex.schema
+    .raw(
+      `
+          create or replace view assignable_campaigns as (
+            select id, title, organization_id, limit_assignment_to_teams
+            from campaign
+            where is_started = true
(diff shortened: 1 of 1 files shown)


bchrobot (changes requested):
Required: run `update-dump.sh` to update `schema-dump.sql`

schema-dump.sql
```diff
@@ -1520,9 +1520,10 @@ CREATE VIEW public.assignable_campaigns AS
  SELECT campaign.id,
     campaign.title,
     campaign.organization_id,
-    campaign.limit_assignment_to_teams
+    campaign.limit_assignment_to_teams,
+    campaign.autosend_status
    FROM public.campaign
-  WHERE ((campaign.is_started = true) AND (campaign.is_archived = false) AND (campaign.is_autoassign_enabled = true));
+  WHERE ((campaign.is_started = true) AND (campaign.is_archived = false) AND (campaign.is_autoassign_enabled = true) AND ((campaign.texting_hours_end)::double precision > date_part('hour'::text, timezone(campaign.timezone, CURRENT_TIMESTAMP))));
```
bchrobot, line 1526: Question: it looks like this got moved up from `assignable_campaigns_with_needs_message`?

Is this something Postgres did under the hood or were there maybe some changes to the migrations file after updating `schema-dump.sql`?

ajohn25: I think it's a mistake because I originally copied from an older migrations file where the clauses were switched - working on fixing!

ajohn25:
@bchrobot think this is all good now but just re-requesting review for a gut check!

migrations/20220517192205_exclude-autosend-campaigns-from-autoassign.js
```diff
@@ -0,0 +1,70 @@
...
+        );
+      `)
+    );
+};
+
+exports.down = function down(knex) {
+  return knex.schema
+    .raw(
+      `
+        create or replace view assignable_campaigns as (
```
bchrobot, line 41: `CREATE OR REPLACE VIEW` [cannot remove columns](https://www.postgresql.org/docs/current/sql-createview.html) so the down migrations will need to drop and recreate the view instead. And that may mean you need to drop and recreate everything down the line.

The up migrations from #1211 might be helpful

bchrobot (changes requested):

migrations/20220517192205_exclude-autosend-campaigns-from-autoassign.js
ajohn25: I think it's just the 2 `assignable_campaigns_with_needs` (message/reply) views that depend on it - checked that the down migrations will actually run this time 🙂

bchrobot (approved):
Looks good!

Issue #1181, opened 2022-05-10
Support template campaigns


bchrobot:
**Is your feature request related to a problem? Please describe.**

Creating many similarly structured campaigns can be tedious.

**Describe the solution you'd like**

Introduction of "Campaign Templates" (or Template Campaigns) and a "New from Template" option.

- Could be implemented as an `is_template` flag on the `campaign` table
- Would live on their own admin page
- Would not show up on the campaign page
- Would not show up in Message Review
- Could not have contacts uploaded
- Could not be assigned to texters
- No campaign actions!

Unit tests should be added for the above ^^.

Templates can be created by:

- Create Blank Template button
- Create Template from Campaign button

In campaign builder page for template, not all sections would be necessary. The following sections could be cut:

- contacts
- filter landlines
- contact overlap management
- texters
- autoassign section (enabled + auto release options)
- texting hours

**Describe alternatives you've considered**

Just using Copy Campaign.

**Additional context**

Use view to "hide" templates:

- rename `campaign` -> `all_campaign`
- `create view campaign as select * from all_campaign where is_template = false`

Design questions:

- [ ] Combine Campaign Groups with Campaign Templates on left sidebar? What would that be called? "Campaign Management" "Campaign Workflows" "Synergy" "Program Management" (separate PR - probably)

bchrobot:
Create from Campaign should have a `quantity` option (campaign title variable `#` for number?)

PR #1211, opened 2022-05-16
chore(template-campaigns): add schema changes

bchrobot:
## Description

This adds schema changes to support template campaigns.

This also includes a fix for `assignable_campaign_contacts`, which was never updated after the tag -> all_tag transition, as a side effect.

## Motivation and Context

Part 1 of _n_ for #1181.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/migrations/20220516035046_template-campaigns.js
+++ b/migrations/20220516035046_template-campaigns.js
@@ -6,31 +6,31 @@ exports.up = function up(knex) {
 
     create view campaign as
       select
-          id
-        , organization_id
-        , title
-        , description
-        , is_started
-        , due_by
-        , created_at
-        , is_archived
-        , use_dynamic_assignment
-        , logo_image_url
-        , intro_html
-        , primary_color
-        , texting_hours_start
-        , texting_hours_end
-        , timezone
-        , creator_id
-        , is_autoassign_enabled
-        , limit_assignment_to_teams
-        , updated_at
-        , replies_stale_after_minutes
-        , landlines_filtered
-        , external_system_id
-        , is_approved
-        , autosend_status
-        , autosend_user_id
+        id,
+        organization_id,
+        title,
+        description,
+        is_started,
+        due_by,
+        created_at,
+        is_archived,
+        use_dynamic_assignment,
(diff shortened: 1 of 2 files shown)


migrations/20220516035046_template-campaigns.js
```diff
@@ -0,0 +1,318 @@
...
+        and is_autoassign_enabled = true
+    );
+
+    create or replace view assignable_needs_message as (
+      select acc.id, acc.campaign_id, acc.message_status
+      from assignable_campaign_contacts as acc
+      join campaign on campaign.id = acc.campaign_id
+      where message_status = 'needsMessage'
+        and (
+            ( acc.contact_timezone is null
```
hiemanshu, line 51: Suggestion: Create `assignable_campaign_within_sending_time` or similar, so when new views are going to be created, they can extend that view instead of having to add the within sending time logic.

hiemanshu (approved):
Overall LGTM! Just a couple of small changes. Feel free to merge/re-request a review after changes.

ajohn25 (approved):

migrations/20220516035046_template-campaigns.js
```diff
@@ -0,0 +1,318 @@
+exports.up = function up(knex) {
+  return knex.schema.raw(`
+    alter table campaign rename to all_campaign;
+
+    alter table all_campaign add column is_template boolean not null default false;
+
+    create view campaign as
+      select
+          id
+        , organization_id
```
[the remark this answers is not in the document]
bchrobot: Done!

migrations/20220516035046_template-campaigns.js
bchrobot: I agree!

ben-pr-p:
> @ben-pr-p curious if you think any changes to indexes are necessary here.

I don't think anyone (even Bernie 2020) has put enough campaigns in Spoke that PostgreSQL would use an index on `is_template_campaign` or would care if any of the other indexes on campaign are partial, so I think we're all good!

PR #1212, opened 2022-05-17
fix(session): handle suspension check when no user is found

bchrobot:
## Description

Check `is_suspended` in a safe way.

## Motivation and Context

Evaluating `user.is_suspended` throws a TypeError when user is undefined.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/auth-passport/local.ts
+++ b/src/server/auth-passport/local.ts
@@ -58,7 +58,7 @@ export function setupLocalAuthPassport() {
           uuidMatch,
           reqBody: req.body
         });
-        if (user.is_suspended) {
+        if (user?.is_suspended === true) {
           return done(new SuspendedUserError());
         }
         return done(null, user);
--- a/src/server/models/cacheable_queries/user.ts
+++ b/src/server/models/cacheable_queries/user.ts
@@ -41,7 +41,7 @@ export async function userLoggedIn(
       ? await getUserByAuth0Id({ auth0Id: val })
       : null;
 
-  return result.is_suspended ? null : result;
+  return result?.is_suspended === true ? null : result;
 }
 
 export async function currentEditors(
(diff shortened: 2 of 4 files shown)


src/server/auth-passport/local.ts
```diff
@@ -58,7 +58,7 @@ export function setupLocalAuthPassport() {
           uuidMatch,
           reqBody: req.body
         });
-        if (user.is_suspended) {
+        if (user?.is_suspended === true) {
```
ajohn25, line 61: Question: Do we need the `=== true` here? If  `user` evaluates to not null then `is_suspended` also should?
```suggestion
        if (user?.is_suspended) {
```

src/server/models/cacheable_queries/user.ts
```diff
@@ -41,7 +41,7 @@ export async function userLoggedIn(
       ? await getUserByAuth0Id({ auth0Id: val })
       : null;
 
-  return result.is_suspended ? null : result;
+  return result?.is_suspended === true ? null : result;
```
ajohn25, line 44: See previous comment

ajohn25 (approved):
1 possible cleanup question but looks good!

src/server/auth-passport/local.ts
bchrobot: A check for truthiness is probably fine here, yes. I wanted to be very explicit though given that it's a pretty critical code path and that Javascript truthy/falsy [can easily bite you](https://www.sitepoint.com/javascript-truthy-falsy/).

hiemanshu: I agree ^^ Suggestion: Add it to the other places above and below too!

hiemanshu (approved):

PR #1222, opened 2022-05-20
fix: require returning non null campaign stats

ajohn25:
## Description
This changes the campaign stats type to have only non null fields.
## Motivation and Context
All of these stats should be valid numbers and not null.

## How Has This Been Tested?
This has been tested locally.
## Checklist:


bchrobot (approved):
👍

Issue #1184, opened 2022-05-10
Campaign variables


bchrobot:
**Is your feature request related to a problem? Please describe.**

There are many times when a campaign is cloned many times with only a few key changes made. It is tedious and error prone to go through and make the edits in every script version of every interaction step.

**Describe the solution you'd like**

Adding campaign-level variables would allow consolidating those few key changes into a variable that could be updated in one place only. This would be a new section in the campaign builder where variable names and values would be defined. Those variables could then be used in scripts and canned responses alongside custom fields.

These should be included in the campaign export. TBD how that store/retrieve campaign variable values _at the time the message was sent_. Options are:

1. Store in `message` table (already a pretty wide table)
2. Only allow soft deletes of campaign variables; add a view that grabs the campaign variable values _at the time the message was sent_.

**Describe alternatives you've considered**

"Variable Sets" -- groups of variables defined as a set that could be attached to multiple campaigns.

**Additional context**

Design questions:

- [x] What happens if campaign variable name conflicts with custom field name?
    - If contacts are already uploaded, prevent saving conflicting campaign variable
    - If contacts are not uploaded, allow upload, show warning on campaign variable section, prevent starting
    - Disallow firstName, lastName, etc.
    - Could also just differentiate (e.g. Zapier)
- [x] Should variable values be preserved when copying a campaign?
    - Generate from template: ignore values, only copy variable names
    - Copy from campaign: copy everything
    - Expand clone campaign/create from template mutation args and modal to add "include variables?" toggle + possibly other options as well

ajohn25:
1. Throw error on start - include variable names
2. Allow interaction step save, stop from starting.

Warning text: "Warning: Variable values are not all present for this script. You can continue working on your script but you cannot start this campaign. The following variables do not have values and will not populate in your script: 
{variable1} 
{variable2}

ajohn25:
@hiemanshu I realized I didn't flag this comment ^ earlier. Our goal is that we're allowed to save scripts with invalid campaign variables but can't start the campaign (we often create the scripts and fill in the variables values later in our use so far). I think #1234 is still blocking saving when you have an invalid one, so would be great to include this change too!

hiemanshu:
@ajohn25 Thanks! I've seen the comment, I'll go back and check on the second scenario, on allowing saving with invalid campaign variables.

ajohn25:
Moved to PR so I could comment more specifically!

PR #1231, opened 2022-05-21
feat: add campaign variable management

bchrobot:
## Description

This adds management of campaign variables.

## Motivation and Context

Part 1 of 2 for #1184.

## How Has This Been Tested?

This has been tested locally and in beta channel releases.

Screen Shot 2022-05-21 at 7 59 38 PM

## Documentation Changes

TBD

## Checklist:

### DIFF
--- a/src/server/api/campaign-variable.ts
+++ b/src/server/api/campaign-variable.ts
@@ -1,11 +1,15 @@
 import { sqlResolvers } from "./lib/utils";
-import { CampaignVariableRecord } from "./types";
 
 export const resolvers = {
   CampaignVariable: {
-    ...sqlResolvers(["id", "name", "value", "createdAt", "updatedAt"]),
-    order: (campaignVariable: CampaignVariableRecord) =>
-      campaignVariable.display_order
+    ...sqlResolvers([
+      "id",
+      "displayOrder",
+      "name",
+      "value",
+      "createdAt",
+      "updatedAt"
+    ])
   }
 };
 
--- a/src/server/api/campaign-variable.ts
+++ b/src/server/api/campaign-variable.ts
@@ -1,0 +1,12 @@
+import { sqlResolvers } from "./lib/utils";
+import { CampaignVariableRecord } from "./types";
+
+export const resolvers = {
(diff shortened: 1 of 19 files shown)


src/server/api/campaign-variable.ts
```diff
@@ -0,0 +1,12 @@
+import { sqlResolvers } from "./lib/utils";
+import { CampaignVariableRecord } from "./types";
+
+export const resolvers = {
+  CampaignVariable: {
+    ...sqlResolvers(["id", "name", "value", "createdAt", "updatedAt"]),
+    order: (campaignVariable: CampaignVariableRecord) =>
```
ajohn25, line 7: Question: Why couldn't this field be called `display_order` throughout?

src/server/api/lib/campaign.ts
```diff
@@ -274,6 +275,27 @@ export const copyCampaign = async (options: CopyCampaignOptions) => {
...
+      // Copy Campaign Variables
+      await trx.raw(
+        `
+          insert into campaign_variable (campaign_id, display_order, name, value)
+          select
+            ? as campaign_id,
+            display_order,
+            campaign_variable.name,
+            (case
+              when all_campaign.is_template then null
```
ajohn25, line 287: Suggestion: We should create an issue for disabling input of a campaign variable value on the template campaign builder page.

ajohn25 (approved):
Exciting!

henryk1229 (approved):

src/server/api/campaign-variable.ts
bchrobot: It definitely could! That's definitely cleaner and one less thing to keep track of. I will change.

I wanted to just call it `order` throughout but that's a SQL reserved word so we would have had to refer to it using `"order"` with the double quotes. Not really a problem, just another thing to remember (but so is order -> display_order).

src/server/api/lib/campaign.ts
bchrobot: Added to this PR!

PR #1234, opened 2022-05-23
feat: support campaign variables in scripts

### LINKED ISSUE #1184: Campaign variables

bchrobot:
## Description

This adds support for using campaign variables in scripts. As part of this, it also stores the ID of the exact version of the campaign variables used in the message.

## Motivation and Context

Closes #1184.

## How Has This Been Tested?

This has been tested locally and in limited beta releases.

<table><tr><td>

Screen Shot 2022-05-23 at 6 38 35 AM

</td></tr><tr><td>

Screen Shot 2022-05-23 at 6 38 47 AM

</td></tr></table>

## Documentation Changes

Documentation has not yet been written.

## Checklist:


ajohn25:
I think we do want the prefixes in! I can take over for that piece?

hiemanshu:
Adding in some more screenshots! And changing this to ready to review! 

Screenshot 2022-07-03 at 4 48 21 PM
Screenshot 2022-07-03 at 4 48 29 PM

Screenshot 2022-07-03 at 4 48 14 PM

src/components/ScriptOptionBlock.tsx
```diff
@@ -105,6 +121,12 @@ export const ScriptOptionBlock: React.FC<ScriptOptionBlockProps> = (props) => {
       {emptyScript && (
         <div className={classes.errorLabel}>Script cannot be empty</div>
       )}
+      {invalidCampaignVariablesUsed.length > 0 && (
+        <div className={classes.errorLabel}>
+          Script cannot use a campaign variable without a variable:{" "}
```
ajohn25, line 126: Nitpick: I think this is probably what was meant?
```suggestion
          Script cannot use a campaign variable without a value:{" "}
```

ajohn25 (approved):

src/containers/AdminCampaignEdit/sections/CampaignInteractionStepsForm/index.tsx
```diff
@@ -284,8 +297,24 @@ const CampaignInteractionStepsForm: React.FC<InnerProps> = (props) => {
...
+      result = result || invalidCampaignVariablesUsed.length > 0;
+    }
+    return result;
+  }, false);
+
   const isSaveDisabled =
-    isWorking || hasEmptyScripts || (!isNew && !hasPendingChanges);
+    isWorking ||
+    hasEmptyScripts ||
+    (!isTemplate && hasInvalidFields) ||
```
ajohn25, line 316: Disabling save here (for non template campaigns)

src/containers/AdminCampaignEdit/sections/CampaignInteractionStepsForm/index.tsx
```diff
@@ -360,6 +410,11 @@ const CampaignInteractionStepsForm: React.FC<InnerProps> = (props) => {
       {hasEmptyScripts && (
         <p style={{ color: "#DD0000" }}>You have one or more empty scripts!</p>
       )}
+      {!isTemplate && hasInvalidFields && (
+        <p style={{ color: "#DD0000" }}>
+          You have one or more scripts using variables without values!
+        </p>
```
ajohn25, line 416: And bringing up this warning

ajohn25 (commented):
It seems like save is still disabled if a variable without a value is used - commenting on lines I think may be responsible below
image

src/server/api/lib/campaign.ts
```diff
@@ -707,3 +707,36 @@ export const editCampaign = async (
...
+  const parsedVariables = variables
+    .map((v: { variable: string }) => v.variable)
+    .flat();
+
+  const campaignContact = await r
+    .knex("campaign_contact")
+    .where({ message_status: "needsMessage", campaign_id: campaignId })
+    .orderBy("updated_at")
+    .first();
+  const customFields = Object.keys(JSON.parse(campaignContact.custom_fields));
```
ajohn25, line 729: Required: Seems like this will error if the campaign has no contacts loaded (seen when trying to create a new campaign).

ajohn25 (changes requested):

ajohn25 (approved):

schema-dump.sql
```diff
@@ -2638,6 +2638,7 @@ CREATE TABLE public.message (
     error_codes text[],
     num_segments smallint,
     num_media smallint,
+    campaign_variable_ids integer[] DEFAULT '{}'::integer[] NOT NULL,
```
bchrobot, line 2641: Required: I think schema-dump.sql needs to be updated again with the 
new `check` for `cv:` prefix.

src/api/campaign.ts
```diff
@@ -229,6 +230,7 @@ export const schema = `
     texters: [User]
     assignments(assignmentsFilter: AssignmentsFilter): [Assignment]
     interactionSteps: [InteractionStep]
+    invalidScriptFields: [String!]
```
bchrobot, line 233: Question: should this be required? or is it actually nullable?

```suggestion
    invalidScriptFields: [String!]!
```

migrations/20220701131714_change_name_constraint_campaign_variables.js
```diff
@@ -0,0 +1,27 @@
+exports.up = function up(knex) {
```
bchrobot, line 1: Required: schema dump needs to be updated with this.

src/containers/AdminCampaignEdit/sections/CampaignVariablesForm.tsx
```diff
@@ -126,7 +126,13 @@ const CampaignVariablesForm: React.FC<FullComponentProps> = (props) => {
     async (formValues: FormValues) => {
       const payload = formValues.campaignVariables
         .filter(({ name }) => !!name)
-        .map((variable, index) => ({ ...variable, displayOrder: index }));
+        .map((variable, index) => {
+          console.log(variable);
```
bchrobot, line 130: Required: remove log

src/api/campaign.ts
hiemanshu: It is actually nullable, because you can have no script fields that are invalid in a campaign.

hiemanshu:
@bchrobot changes made! Let me know if this is good to go!

src/api/campaign.ts
bchrobot: I think returning an empty array makes more sense in that case

hiemanshu: It does return an empty array, I'll fix the type def!

PR #1253, opened 2022-06-03
fix(external-system): fix saving credentials

bchrobot:
## Description

This fixes updating credentials for an external system when the underlying graphile secret somehow does not exist.

## Motivation and Context

graphile-secrets throws an error when the requested secret isn't found. Spoke expects an undefined value.

Ref: https://secure.helpscout.net/conversation/1905190944/6480/

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -3048,9 +3048,9 @@ const rootMutations = {
         username: externalSystem.username
       };
 
-      const savedSystemApiKeySecret = await getWorker().then((worker) => {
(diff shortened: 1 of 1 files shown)


ajohn25 (approved):

hiemanshu (approved):
LGTM! Seems like we fixing the symptoms rather than the root cause here, do we know why graphile secret may be undefined?

PR #1255, opened 2022-06-03
fix(autosending): preserve run_at when queuing messages

### LINKED ISSUE #1246: Remove overwrite of autosend run time when queuing
**Describe the bug**
In [`queue-autosend-initials.ts`](https://github.com/politics-rewired/Spoke/blob/main/src/server/tasks/queue-autosend-initials.ts#L94), every time the job runs, the `run_at` column is overwritten. In some cases, this leads to autosending for affected contacts being pushed indefinitely into the future.

**To Reproduce**
Steps to reproduce the behavior: (not confirmed yet)
1.  Start autosending on a large campaign (>10k contacts)
2.  Go to Message Review and reassign a large number of initials to a user (we expect that this step causes [`queue-autosend-initials.ts`](https://github.com/politics-rewired/Spoke/blob/main/src/server/tasks/queue-autosend-initials.ts#L94) to run again, and overwrite the `run_at` time.
3.  See that messages for the campaign stop sending

**Expected behavior**
Messages should continue sending unless an autosend campaign is paused.

bchrobot:
## Description

This preserves `run_at` for autosend message tasks so they cannot continually be pushed into the future.

## Motivation and Context

On conflict approach taken from https://stackoverflow.com/a/42217872

`UNION` does not work with JSON so this also limits the returned table to only required columns.

Closes #1246

## How Has This Been Tested?

This has been tested locally to ensure result set is the same and `run_at` is not modified.

N/A

## Documentation Changes

N/A

## Checklist:

---
  - https://app.asana.com/0/0/1202377870515168

### DIFF
--- a/src/server/tasks/queue-autosend-initials.ts
+++ b/src/server/tasks/queue-autosend-initials.ts
@@ -73,7 +73,7 @@ const queueAutoSendInitials: Task = async (payload: Payload, helpers) => {
         where id in ( select id from contacts_to_queue )
         returning id
       ),
-      jobs_queued as (
+      new_jobs_queued as (
         insert into graphile_worker.jobs (task_identifier, payload, key, queue_name, max_attempts, run_at)
         select 
           'retry-interaction-step' as task_identifier, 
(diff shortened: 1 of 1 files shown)


ben-pr-p (approved):
I think this works, but maybe its simpler to do `run_at = least(graphile_worker.jobs.run_at, excluded.run_at)` so there's always an update?

This is good too though

ajohn25 (approved):

PR #1256, opened 2022-06-04
fix: fix post-login redirect behavior for local auth

bchrobot:
## Description

This fixes post-login redirect behavior for local passport strategy.

## Motivation and Context

The standardized post-login redirect handler only looks at state preserved during oauth flow, not the nextUrl query param.

Local auth login is an async client request so a redirect response doesn’t even make sense as the way to handle successful login. Instead we can just use browser history.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/containers/UserEdit.jsx
+++ b/src/containers/UserEdit.jsx
@@ -12,6 +12,8 @@ import PropTypes from "prop-types";
 import queryString from "query-string";
 import React from "react";
 import Form from "react-formal";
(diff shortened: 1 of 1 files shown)


hiemanshu (approved):
Ah. Thanks for fixing this :D

ajohn25 (approved):

Issue #1220, opened 2022-05-20
Notify user when they change their information


ajohn25:
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**
Saving user information notifies the user that they have changed their notification preferences.

**To Reproduce**

1.  Click on user info from the top right menu
2.  Change the first name
3.  Click save
4.  See the notification that the user's notification frequency has been updated

**Expected behavior**
A notification that user information has been updated

**Screenshots**
![image](https://user-images.githubusercontent.com/76596635/169444648-e70bd230-c260-4bdb-99ec-125ef4c13705.png)

**Additional context**
This could probably be achieved by just removing the notification frequency specific text if the change is not for notification frequency. The top of the notification which just says "Changes saved" is enough to indicate that the save succeeded.

PR #1257, opened 2022-06-04
fix: show notification settings alert only on user edit

bchrobot:
## Description

This changes the notification frequency save  result alert to only show after an edit user operation.

## Motivation and Context

The alert was showing after all local auth methods, including those where notification settings are not touched (eg login).

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:


ajohn25 (approved):
If you wanted to, you could also check that the edit is for notification type and not other user info to close #1220, but that could be done separately too.

hiemanshu (approved):

Issue #1232, opened 2022-05-22
Support per-campaign messaging service selection


bchrobot:
### Is your feature request related to a problem? Please describe.

Currently, Spoke _technically_ supports multiple messaging services per organization. In practice, only one messaging service per organization is supported, however.

This means that managing different sending number pools (e.g. toll-free vs long code) requires users to fragment their work across multiple Spoke organizations in order to use multiple number pools. This makes it very difficult to transition traffic in case of severe outage.

### Describe the solution you'd like

Spoke should support multiple messaging services per organization and selection of messaging services per-camapign within an organization.

### Describe alternatives you've considered

N/A

### Additional context

**Campaign Builder**

When building a campaign, messaging service selection should only be shown if there is more than one active messaging service. Otherwise the single active messaging service should be used by default and the UI for selection should be hidden.

**messaging_service**

Messaging Services will need the following new properties:

- `name text` -- user-friendly display name
- `is_active boolean` -- mark a messaging service as inactive. This would prevent any new campaigns from using the messaging service but would still handle inbound replies for existing campaigns.

**messaging_service_stick**

This table can be deprecated as the mapping will live at the campaign level, rather than the phone number level (a holdover from supporting migrating messaging services based on campaign_contact record data).

**Telco webhook handling**

Reply handling will need to be updated to use the mapping stored on `campaign`:

https://github.com/politics-rewired/Spoke/blob/fc665424064d68585abb42c7f5d23f8c3014d74e/src/server/api/lib/message-sending.js#L313-L348

Delivery report handling shouldn't required any changes.

bchrobot:
For backwards compatibility, the new `campaign.messaging_service_id` column could be nullable and messaging changes could switch based on whether that's null or not: if null, use messaging_service_stick, otherwise use new per-campaign logic.

bchrobot:
This can be broken into 3 sub-tasks:

1. The migrations and backend changes outlined above. This allows us to begin using this feature for emergency migration of traffic.
    - Campaign creation would use the single active messaging service and throw an error if there are either 0 or more than 1 active messaging services.
2. Campaign builder UI and mutation changes to support selecting from a list of active messaging services. This would allow clients to take advantage of the feature, although messaging services would still need to be added manually.
3. Add messaging service management to client app (low priority)

hiemanshu:
Specing this out a little more before I start work on it. 

Part 1:

1. Update messaging service table to add user friendly name, and active field
[...]

PR #1260, opened 2022-06-06
feat: allow campaign to pick messaging service

hiemanshu:
## Description

Part 1 of 3 allowing campaigns to pick their own messaging service for that campaign

## Motivation and Context

Part 1 of #1232 

## How Has This Been Tested?

Tested Locally.

## Checklist:

### DIFF
--- a/migrations/20220605033846_add_messaging_service_to_campaign.js
+++ b/migrations/20220605033846_add_messaging_service_to_campaign.js
@@ -1,0 +1,84 @@
+exports.up = function up(knex) {
+  return knex.schema
+    .alterTable("all_campaign", (table) => {
+      table.text("messaging_service_sid");
+
+      table
+        .foreign("messaging_service_sid")
+        .references("messaging_service.messaging_service_sid");
+    })
+    .then(() => {
+      return knex.schema.raw(`
+    create or replace view campaign as
+      select
+        id,
+        organization_id,
+        title,
+        description,
+        is_started,
+        due_by,
+        created_at,
+        is_archived,
+        use_dynamic_assignment,
+        logo_image_url,
+        intro_html,
+        primary_color,
+        texting_hours_start,
+        texting_hours_end,
+        timezone,
+        creator_id,
+        is_autoassign_enabled,
+        limit_assignment_to_teams,
+        updated_at,
+        replies_stale_after_minutes,
+        landlines_filtered,
+        external_system_id,
+        is_approved,
+        autosend_status,
+        autosend_user_id,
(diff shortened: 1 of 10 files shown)


migrations/20220605033846_add_messaging_service_to_campaign.js
```diff
@@ -0,0 +1,84 @@
...
+        autosend_status,
+        autosend_user_id,
+        messaging_service_sid
+      from all_campaign
+      where is_template = false;
+`);
+    });
+};
+
+exports.down = function down(knex) {
```
bchrobot, line 46: Question: have you tested the down migration? I think the order of the two needs steps needs to be swapped: removing the column from the view before dropping the underlying column.

You might also need to redefine downstream views (see [1121 migration](https://github.com/politics-rewired/Spoke/pull/1211/files#diff-729cef21459886518148a2a70dadc29b0164b923b26bea0985d33f97ede4ecfe) for example).

bchrobot (changes requested):
Required: `getCampaignContactAndAssignmentForIncomingMessage()` needs to be updated to account for the fact that any given `cell` within an organization can now have been texted by more than one messaging service.

Required: createCampaign and copyCampaign should set `campaign.messaging_service_sid` on creation to the single active messaging service. So we begin using the new code path right away. I believe this is different from part 2 because it is happening entirely behind the scenes and doesn't touch the GraphQL schema at all.

ajohn25 (approved):

src/server/api/lib/campaign.ts
```diff
@@ -203,6 +203,28 @@ export const copyCampaign = async (options: CopyCampaignOptions) => {
         [count, userId, campaignId]
       );
 
+      // Copy Messaging Service OR use active one
```
bchrobot, line 206: I talked with @kohlee about this and we think that `messaging_service_sid` shouldn't copy over. For now, it should always use the one active messaging service or throw an error

src/server/api/schema.js
```diff
@@ -802,6 +802,11 @@ const rootMutations = {
         features
       );
 
+      const messagingService = await r
+        .knex("messaging_service")
+        .where({ organization_id: campaign.organizationId, active: true })
+        .first();
```
bchrobot, line 808: For now, this should return all rows `where active = true` and error if `length !== 0`. We can revisit if/when we start exposing messaging service management to users.

bchrobot (changes requested):

bchrobot (approved):
Looks good!

PR #1263, opened 2022-06-08
feat: add bandwidth error codes

### LINKED ISSUE #1249: Add Bandwidth error codes to Spoke error code mapping
Now that we're also sending from Bandwidth, we need error codes to be friendly in the UI.

Should just be adding new things here: https://github.com/politics-rewired/Spoke/blob/7931f4c0587bb3e5b2927a314dcc538c70ce74b0/src/containers/AdminCampaignStats/DeliverabilityStats.tsx#L12

From here: https://dev.bandwidth.com/docs/messaging/errors/

ben-pr-p:
## Description

Add Bandwidth error codes from the docs here: https://dev.bandwidth.com/docs/messaging/errors/

In general, I tried to simplify them were complicated while preserving what the user needs to know.

## Motivation and Context

Closes #1249 

## How Has This Been Tested?

N/A

## Documentation Changes

Change made at the bottom here: https://docs.spokerewired.com/article/90-send-status-and-error-codes

## Checklist:

- [x] I have included updates for the documentation accordingly.

### DIFF
--- a/src/containers/AdminCampaignStats/DeliverabilityStats.tsx
+++ b/src/containers/AdminCampaignStats/DeliverabilityStats.tsx
@@ -29,7 +29,22 @@ const descriptions: Record<string, string> = {
   "30005": "Unknown destination handset",
   "30006": "Landline or unreachable carrier",
   "30007": "Blocked as spam",
-  "30008": "Unknown error"
+  "30008": "Unknown error",
+  "4405": "Unreachable sending phone number", // this one shouldn't happen
+  "4406": "Unreachable phone number",
+  "4432": "Unreachable country",
+  "4434": "Unreachable toll-free number",
+  "4700": "Landline or unreachable carrier",
+  "4720": "Unreachable phone number",
+  "4730": "Unreachable phone number (recipient maybe roaming)",
+  "4740": "Unreachable sending phone number", // this one shouldn't happen, but has happening
+  "4750": "Carrier rejected (maybe spam)",
+  "4770": "Blocked as spam",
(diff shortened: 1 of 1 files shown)


src/containers/AdminCampaignStats/DeliverabilityStats.tsx
```diff
@@ -29,7 +29,22 @@ const descriptions: Record<string, string> = {
...
-  "30008": "Unknown error"
+  "30008": "Unknown error",
+  "4405": "Unreachable sending phone number", // this one shouldn't happen
+  "4406": "Unreachable phone number",
+  "4432": "Unreachable country",
+  "4434": "Unreachable toll-free number",
+  "4700": "Landline or unreachable carrier",
+  "4720": "Unreachable phone number",
+  "4730": "Unreachable phone number (recipient maybe roaming)",
+  "4740": "Unreachable sending phone number", // this one shouldn't happen, but has happening
```
bchrobot, line 40: Nit: grammar

```suggestion
  "4740": "Unreachable sending phone number", // this one shouldn't happen, but has happened
```

Aside: any idea why? Or follow up tasks?

bchrobot (approved):

src/containers/AdminCampaignStats/DeliverabilityStats.tsx
ben-pr-p: We need to create tickets with Bandwidth probably - it's only happened for a handful of messages, but both indicate something not 100% right with phone number provisioning

PR #1285, opened 2022-06-23
fix: fix copy campaign mutation

ajohn25:
## Description
This fixes setting messaging services when copying a campaign, by using the object that's returned directly.

## Motivation and Context
Creating campaigns from a template did not work after changes to this mutation in #1260 

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/src/server/api/lib/campaign.ts
+++ b/src/server/api/lib/campaign.ts
@@ -204,14 +204,9 @@ export const copyCampaign = async (options: CopyCampaignOptions) => {
       );
 
       // Copy Messaging Service OR use active one
-      const campaign = await r
-        .knex("campaign")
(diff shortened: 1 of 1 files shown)


bchrobot (approved):
Ah, so the issue was the original query was selecting from `campaign` (does not include template campaigns) instead of `all_campaign` (includes template campaigns).

Suggestion: add two other test cases like [this one](https://github.com/politics-rewired/Spoke/blob/main/src/server/api/lib/campaign.spec.ts#L152-L181) for copying a campaign and creating from template. Happy to pair on this if it would be helpful!

hiemanshu (approved):

Issue #826, opened 2021-01-06
Make Message Review filters shareable by url


kohlee:
**Is your feature request related to a problem? Please describe.**
Admins are unable to share Message Review urls with fellow users and have the filters pre-populate with values that were present when the url was copied. 

**Describe the solution you'd like**
Message Review views should be shareable by url. 

**Describe alternatives you've considered**
N/A

**Additional context**
This is supported in MoveOn's fork.

hiemanshu:
@bchrobot For this I was thinking of moving most of the state into a search params of the URL. So the component reads the values from the URL, and any changes in filters cause a change in the URL, this way the URL can just be copy pasted and shared with anyone else. 

I've tried to experiment with a few different ideas, and instead of maintaining multiple sources of truth, moving the source to the URL made the most sense.

hiemanshu:
@bchrobot following up on this, do you have any other ideas on the refactoring?

bchrobot:
## Current State

Message Review currently looks like:

```tsx
<AdminIncomingMessageList>
  {/* Workaround for fetching data */}
  <PaginatedUsersRetriever />
  <PaginatedUsersRetriever />
  <PaginatedCampaignsRetriever />


  <IncomingMessageFilter>
    <Card>
      <Toggle />
      <Toggle />
      <SelectField />
      <Autocomplete />
      <Autocomplete />
      <TextField />
      <SelectField />
    </Card>
  </IncomingMessageFilter>

  <IncomingMessageActions />

  <IncomingMessageList>
    <DataTable />
    <ConversationPreviewModal />
  </IncomingMessageList>

  <ReassignmentResultDialog />
</AdminIncomingMessageList>
```

Drawbacks:

- The top level `<AdminIncomingMessageList>` handles _lots_ of state, business logic, and layout
- The data retrievers are a clunky work around for an outdated mode of using Apollo Client
- `<IncomingMessageFilter>` manages the container/layout and business logic for all filter form input elements


## Proposal

- Extract business logic and state management into a standalone hook using `useReducer()`
    - Campaign filter state stored in URL query params (e.g. `?campaign=6&texter=unassigned`)
    - Conversation detail modal ID stored in URL query param (e.g. `?activeConversation=5`)
- Break up filter component. Container handles layout, individual components handle filter-specific logic (e.g. resolving multiselect dropdown value into a valid GraphQL query input object)
[...]

PR #1288, opened 2022-06-24
feat(message review): use url to store filters

hiemanshu:
Using the url to store the filters allows sharing the link

## Description

Use `useQueryParams` to store the computed filters in the URL. This allows sharing the URL to share the computed filters used to fetch messages in message review.

## Motivation and Context

This is part 5 of 7 for #826 

## How Has This Been Tested?

Tested locally with quite a few different combinations.

## Checklist:

### DIFF
--- a/src/containers/AdminIncomingMessageList/index.tsx
+++ b/src/containers/AdminIncomingMessageList/index.tsx
@@ -11,6 +11,7 @@ import {
 import omit from "lodash/omit";
 import React, { useState } from "react";
 import { useParams } from "react-router-dom";
+import { NumberParam, useQueryParam, withDefault } from "use-query-params";
 
 import { ALL_TEXTERS, UNASSIGNED_TEXTER } from "../../lib/constants";
 import IncomingMessageActions from "./components/IncomingMessageActions";
@@ -20,6 +21,13 @@ import {
   getCampaignsFilterForCampaignArchiveStatus,
   getContactsFilterForConversationOptOutStatus
 } from "./filter-utils";
+import {
+  AssignmentsFilterParam,
+  CampaignsFilterParam,
+  ContactNameParam,
+  ContactsFilterParam,
+  TagsFilterParam
+} from "./types";
 
 /* Initialized as objects to later facillitate shallow comparison */
 const initialCampaignsFilter = { isArchived: false };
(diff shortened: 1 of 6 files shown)


src/containers/AdminIncomingMessageList/types.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    const filter: CampaignsFilter = {};
+
+    for (const [key, value] of Object.entries(decodedObject)) {
+      if (key === "isArchived") {
+        filter[key] = value === "true";
+      } else {
+        filter[key] = parseInt(value, 10);
+      }
```
bchrobot, line 71: Suggestion: Prefer explicit -- use a `switch` with explicit cases for each of the keys.

src/containers/AdminIncomingMessageList/types.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    const contactsFilter: ContactsFilter = {};
+
+    for (const [key, value] of Object.entries(decodedObject)) {
+      if (["isOptedOut", "validTimeZone", "includePastDue"].includes(key)) {
+        contactsFilter[key] = value === "true";
+      } else {
+        contactsFilter[key] = value;
+      }
```
bchrobot, line 34: Suggestion: Prefer explicit -- use a `switch` with explicit cases for each of the keys.

src/containers/AdminIncomingMessageList/types.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    const filter: AssignmentsFilter = {};
+
+    for (const [key, value] of Object.entries(decodedObject)) {
+      if (key === "includeEscalated") {
+        filter[key] = value === "true";
+      } else {
+        filter[key] = parseInt(value, 10);
+      }
```
bchrobot, line 100: Suggestion: Prefer explicit -- use a `switch` with explicit cases for each of the keys.

src/containers/AdminIncomingMessageList/types.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    const filter: TagsFilter = {};
+
+    for (const [key, value] of Object.entries(decodedObject)) {
+      if (["excludeEscalated", "escalatedConvosOnly"].includes(key)) {
+        filter[key] = value === "true";
+      } else if (key === "specificTagIds") {
+        filter[key] = decodeDelimitedArray(value, ",");
+      }
```
bchrobot, line 130: Suggestion: use a `switch` instead of if/else if/else for consistency with above suggestions.

but otherwise I am onboard with complete enumeration of TagsFilter keys

src/containers/AdminIncomingMessageList/index.tsx
```diff
@@ -63,34 +71,56 @@ const AdminIncomingMessageList: React.FC<AdminIncomingMessageListProps> = (
...
+    withDefault(NumberParam, 10)
   );
-  const [tagsFilter, setTagsFilter] = useState<any>(defaultTagsFilter);
-  const [contactNameFilter, setContactNameFilter] = useState<any>(undefined);
+
+  // Get or Compute Default states
+  const defaultArchivedCampaigns = campaignsFilter.isArchived ?? false;
+  const defaultActiveCampaigns = !campaignsFilter.isArchived ?? true;
+  const defaultIncludeNotOptedOut = contactsFilter.isOptedOut ?? false;
+  const defaultIncludeOptedOut = !contactsFilter.isOptedOut ?? true;
```
bchrobot, line 106: Question: how does setting a default based on a dynamic value work with `use-query-params`?

Does it only evaluate it based on the very first value of `contactsFilter`?

Or does it keep it in sync with `contactsFilter` until a user has made a change? If so, does it override the user's selection if `contactsFilter` changes?

bchrobot (approved):
@kohlee is gonna go bananas for this!

bchrobot (approved):
👍

ajohn25 (approved):

src/containers/AdminIncomingMessageList/index.tsx
hiemanshu: It evaluates based on when the component is loaded, since this computes the default value, and then the default value is used in `useState` which doesn't change even when filter changes, but only when we change the state which is done by user toggling. 

useQueryParams will read the URL value, or set a default value that we gave to it (without updating the URLs), so if it reads the URL to find a value, the campaignsFilter will be what the URL is, if not, the campaignsFilter will be the default we've given.

PR #1290, opened 2022-06-24
fix: mark landlines as unfiltered on new contacts upload

### LINKED ISSUE #907: Reset Filter Landlines after new list is loaded
**Describe the bug**
Landline filter status is not reset when second list is loaded

**To Reproduce**

1.  In the campaign builder, go to Contacts and load a contact list
1. Go to Filter Landlines and click Filter Landlines button
1. Go to Contacts and load a new contact list
1. Go to Filter Landlines and see output of step 2

**Expected behavior**
After a new list is loaded, the Filter Landlines section should be reset and no longer display the previous list's filtering result message.

**Desktop**

* OS: iOS
* Browser: Chrome
* Version [e.g. 22]

**Additional context**
Some rethinking of the Filter Landlines section should also be done regarding: 

* the ability to download a list of rejected landlines
* making the result message of the filter operation more user-friendly

hiemanshu:
## Description

The change to landlines being filtered can be slow depending on the load on the server, making the change earlier resets the result.

## Motivation and Context

This fixes #907 

## How Has This Been Tested?
Tested locally

## Checklist:

### DIFF
--- a/src/workers/jobs/index.js
+++ b/src/workers/jobs/index.js
@@ -152,13 +152,7 @@ export async function uploadContacts(job) {
 
   const orgFeatures = JSON.parse(organization.features || "{}");
 
-  await Promise.all([
-    r.knex("campaign_contact").where({ campaign_id: campaignId }).del(),
-    r
-      .knex("campaign")
-      .update({ landlines_filtered: false })
-      .where({ id: campaignId })
-  ]);
+  await r.knex("campaign_contact").where({ campaign_id: campaignId }).del();
 
   let jobPayload = await gunzip(Buffer.from(job.payload, "base64"));
   jobPayload = JSON.parse(jobPayload);
@@ -617,6 +611,11 @@ export async function loadContactsFromDataWarehouse(job) {
     .knex("campaign_contact")
     .where("campaign_id", job.campaign_id)
     .delete();
+
+  await r
+    .knex("campaign")
+    .where({ id: campaign.id })
+    .update({ landlines_filtered: false });
 
   await loadContactsFromDataWarehouseFragment({
     jobId: job.id,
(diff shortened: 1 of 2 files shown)


src/workers/jobs/index.js
```diff
@@ -618,6 +612,11 @@ export async function loadContactsFromDataWarehouse(job) {
     .where("campaign_id", job.campaign_id)
     .delete();
 
+  await r
+    .knex("campaign")
+    .where({ id: campaign.id })
+    .update({ landlines_filtered: false });
```
bchrobot, line 618: Suggestion: move this to editCampaign in `src/server/api/lib/campaign.ts` around line 490 for consistency with the other two changes.

bchrobot (commented):
Question: the root cause is not completely clear to me. 

Is my understanding correct that the database you're load referring to is how long it takes graphile-worker to pick up the `uploadContacts` task? It looks like the update to `landlines_filtered` used to happen at the beginning of `uploadContacts()` in its own (implicit) transaction.

In that case, we could expect the version on `main` to eventually reach the correct state once all jobs complete. I am not sure that is what @kohlee has experienced

hiemanshu:
> Is my understanding correct that the database you're load referring to is how long it takes graphile-worker to pick up the uploadContacts task? It looks like the update to landlines_filtered used to happen at the beginning of uploadContacts() in its own (implicit) transaction.

Yup. The time taken for the task to get picked up could a visual delay. I tested all other scenarios, and locally had no other issues I could find with this. 

@kohlee can you confirm what you've experienced with this issue?

bchrobot (approved):
Approved pending Kohlee's confirmation of the issue.

Also flagging the _suggestion_ from the previous code review in case that gets lost in the comments

ajohn25 (approved):
From organizing team's perspective, the issue is that filter landlines wasn't reset in the past unless a dev manually set `landlines_filtered = false` for the campaign (ex. we couldn't filter again on the same campaign after upload normally). So it _seems_ like not updating landlines_filtered when uploading contacts would be the issue we're focused on vs. delays from graphile-worker. But if visual delay seems like the only behavior to correct, let's merge and QA!

PR #1299, opened 2022-06-30
fix(message-review): listen for clicks on entire card header

bchrobot:
## Description

This adds the onClick listener to the Filters and Actions card headers rather than just the action button.

## Motivation and Context

Users expect to be able to click on the entire header (as was previously the behavior).

Initial work for addressing #1302

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

---
  - https://app.asana.com/0/0/1202532271228785


ajohn25 (approved):
Suggestion: While doing this, can we change [this line](https://github.com/politics-rewired/Spoke/blob/main/src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx#L133) to open the message filter by default? (No one ever opens this screen without wanting to open the filter too)

Issue #1218, opened 2022-05-20
Clear contact message status filter in message review by default


ajohn25:
**Your issue may already be reported!**
Please search on the [issue tracker](../) before creating one.

- [x] I have searched through existing issues and did not find an existing report

**Describe the bug**
The contact message status filter on the message review page has "All" selected when the page loads.

**To Reproduce**

1.  Go to Message Review
2.  Click on the contact message status filter
3.  See that all values are selected

**Expected behavior**
The contact message status filter on the message review page should be cleared when the page loads.

**Screenshots**
![image](https://user-images.githubusercontent.com/76596635/169431970-7b0d4d8c-d113-413e-b259-7f95913a8015.png)

Additional context
Likely introduced by MUI v4 upgrade in #1178.

ajohn25:
@hiemanshu the problems you're working through on #1305 reminded me about this, so just wanted to flag as hopefully a quick fix that saves many people extra clicks if you have time for it!

PR #1305, opened 2022-07-01
fix: set initial values of message filters reading

### LINKED ISSUE #1303: Shared Message Review URL sets GraphQL query params but not Filters UI state
**Describe the bug**

Now that Message Review filter state is stored in query parameters users may share URLs for specific Message Review queries. When opening one of these shared links, the URL query parameters correctly set the state of the GraphQL query to fetch conversations. They do not set initial UI state in the Message Filters component, however, making it difficult for the user to know what they're looking at.

**To Reproduce**

1.  Go to Message Review
2.  Change some filter options in Message Filters
3.  Copy the URL
4.  Open the URL in a new tab
5. See that conversations GraphQL query runs with expected parameters
6. See that Message Filters in the new tab does not reflect query parameters

**Expected behavior**

Opening a Message Review link with query parameters should set initial UI state based on the URL query parameters.

**Screenshots**

Screen_Shot_2022-06-30_at_8_29_55_AM

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Brave
- Version: Version 1.37.116 Chromium: 100.0.4896.127 (Official Build) (x86_64)

**Additional context**

Two thoughts on solutions for this are:

1. Convert `<IncomingMessageFilter />` to a controlled component. For `value` props on filter inputs, replace values from `useState()` with values from props. Continue using `onChange` props.
2. Treat the query params as an external state store. Replace `useState` and `props.onChange`s in `<IncomingMessageFilter />` with `useQueryParam`.

The two `<Autocomplete />`s will be tricky. The ID of the selected campaign or user will be stored in the URL but in order to display correctly, the appropriate record needs to be present in the Autocomplete's `options`. The `options` are sourced from a GraphQL query that relies on TextField state stored locally.

hiemanshu:
## Description

Read and setup initial state of filters

## Motivation and Context

Fixes  #1303 

## How Has This Been Tested?

Tested Locally

Screenshot 2022-07-01 at 5 03 57 PM
Screenshot 2022-07-01 at 5 04 25 PM

## Checklist:


ajohn25 (approved):

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
```diff
@@ -449,6 +487,7 @@ const IncomingMessageFilter: React.FC<IncomingMessageFilterProps> = (props) => {
           <TextField
             className={css(styles.fullWidth)}
             onChange={onContactNameChanged}
+            defaultValue={contactFilterDefault}
```
bchrobot, line 490: Suggestion: make this a controlled component as well

```suggestion
            value={contactFilter}
```

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
```diff
@@ -360,6 +384,14 @@ const IncomingMessageFilter: React.FC<IncomingMessageFilterProps> = (props) => {
     );
   };
 
+  let contactFilterDefault = firstName ?? "";
+  if (lastName) {
+    contactFilterDefault = contactFilterDefault.concat(" ", lastName);
+  }
+  if (cellNumber) {
+    contactFilterDefault = contactFilterDefault.concat(" ", cellNumber);
+  }
```
bchrobot, line 393: Suggestion: use this to make the TextField a controlled component (see other comment)

```suggestion
  let contactFilter = firstName ?? "";
  if (lastName) {
    contactFilter = contactFilter.concat(" ", lastName);
  }
  if (cellNumber) {
    contactFilter = contactFilter.concat(" ", cellNumber);
  }
```

Suggestion/nit: this could also be a place to use `useMemo()`

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
```diff
@@ -202,6 +211,19 @@ const IncomingMessageFilter: React.FC<IncomingMessageFilterProps> = (props) => {
...
+      setCampaignOption(
+        campaignOptions.find((campaign) => campaign.id === props.campaignId)
+      );
+    }
+    if (!isEmpty(texterOptions) && props.texterId) {
+      setTexterOption(
+        texterOptions.find((texter) => texter.id === props.texterId)
+      );
+    }
+  }, [getCampaigns, getTexters]);
```
bchrobot, line 225: Suggestion: use Autocomplete's `getOptionSelected()` prop to support using the `campaignId` and `texterId` props for the Autocomplete's `value` directly, instead of setting the value to be the matching `option`.

From the [API reference](https://v4.mui.com/api/autocomplete/)
> `value` -- The value must have reference equality with the option in order to be selected. You can customize the equality behavior with the `getOptionSelected` prop.

bchrobot (approved):
Looks good!

Just a few suggestions for further simplifying state management.

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
hiemanshu: Had to go back on this too. We use a function to trim the name components, and trying to add a space after the first word would not happen because the spaces would get trimmed. I can make it controlled by setting the `name` as a full value rather than the individual `firstName, lastName, cellNumber`, but that complicates things some more.

hiemanshu:
Setting a default value seems to causing some unintended effects when the default value isn't actually passed by the contactsFilter, going to go back and think some more about this. It selects `All Campaigns` by default, which removes all of the campaign options because of the way the autocomplete is now setup to search with the text.

hiemanshu:
I made a bunch of large changes, so getting a second review. This also fixes #1307 and #1218

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
bchrobot: Ahh that is so frustrating! I just looked at the type definition and that is indeed how it is implemented 😒

Option and value could just be campaign IDs then, with `getOptionLabel()` used to get the title. The current `options` array could be reduced to a map from `campaignId` to label so the lookups would be fast.

bchrobot (approved):

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
bchrobot: Got it.

Maybe a follow-up issue that does two things:
1. Moves the query string splitting and mapping to Postgres where clauses to the backend.
2. Makes this a controlled component with just a query string.

What do you think?

src/containers/AdminIncomingMessageList/components/IncomingMessageFilter.tsx
hiemanshu: If we used just Campaign IDs, we’d have to use an Array for the options and a map for the lookup. Should be easy to implement. I’ll see if there are any other issues while trying to implement it, otherwise will go ahead and implement that.

bchrobot (approved):

ajohn25 (approved):

PR #1315, opened 2022-07-07
fix(campaign builder): return nullable messaging services

ajohn25:
## Description
This changes the messaging services object to be null when the user is not an owner. This prevents a "FORBIDDEN" GraphQL error from being thrown.
## Motivation and Context
Users who were not owners couldn't load the campaign edit page, due to the "FORBIDDEN" GraphQL error that was in the response.

## How Has This Been Tested?
This has been tested locally.
## Checklist:

### DIFF
--- a/src/api/organization.ts
+++ b/src/api/organization.ts
@@ -112,7 +112,7 @@ export const schema = `
     escalationTagList: [Tag]
     teams: [Team]!
     externalSystems(after: Cursor, first: Int): ExternalSystemPage!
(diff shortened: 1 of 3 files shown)


bchrobot (approved):
Looks good to me!

Only comment would be to create a follow-up issue to revert the try/catch change after Message Review is eventually refactored to replace our `loadData()` HOC with Apollo's `useQuery()` hook. The GraphQL property still needs  nullable but `useQuery()` will let us choose how we handle the permission error rather than forcing the display of an error message.

hiemanshu (approved):

PR #1327, opened 2022-07-23
fix: fix saving empty team priority

ajohn25:
## Description
This prevents saving a team with a null priority value, and also adjusts the GraphQL schema to allow retrieving teams which have null priority values.

## Motivation and Context
Due to the visual bug on the create team form that presents an empty string value as 500 to the user, the user can create a team with a null priority. If a team with a null priority exists, loading the teams page isn't possible, because the GraphQL schema expects non null assignment priority values.

We generally want teams to have assignment priorities, but there are a number of database functions that already expect this column to be nullable. Assigning default priorities for team creation, and allowing the loading of teams if the priority value is null feels like a good solution here for these reasons.

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/src/containers/AdminTeamEditor/index.jsx
+++ b/src/containers/AdminTeamEditor/index.jsx
@@ -64,7 +64,9 @@ class AdminTeamEditor extends Component {
       "assignmentPriority"
     ]);
     if ("assignmentPriority" in team) {
-      team.assignmentPriority = parseInt(team.assignmentPriority, 10);
+      const rawPriority = team.assignmentPriority;
+      team.assignmentPriority =
+        rawPriority === "" ? 500 : parseInt(team.assignmentPriority, 10);
     }
 
     this.setState({ isWorking: true });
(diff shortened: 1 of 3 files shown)


bchrobot (approved):
Nice!

src/containers/AdminTeamEditor/index.jsx
```diff
@@ -64,7 +64,9 @@ class AdminTeamEditor extends Component {
       "assignmentPriority"
     ]);
     if ("assignmentPriority" in team) {
-      team.assignmentPriority = parseInt(team.assignmentPriority, 10);
+      const rawPriority = team.assignmentPriority;
+      team.assignmentPriority =
+        rawPriority === "" ? 500 : parseInt(team.assignmentPriority, 10);
```
hiemanshu, line 69: nitpick: parseInt for `rawPriority` since we're already reading that value.

hiemanshu (approved):

src/containers/AdminTeamEditor/index.jsx
ajohn25: Thanks, whole reason I wrote that 1st line and still forgot to change that lol

Issue #783, opened 2020-11-17
Add superadmin page


bchrobot:
**Is your feature request related to a problem? Please describe.**

There is no way for instance admins (superadmins) to:

1. Manage users that do not belong to an organization
1. Manage other superadmins
1. View billing reports

**Describe the solution you'd like**

Superadmins should have access to a superadmin portal on their instance to manage these things.

**Describe alternatives you've considered**

N/A

**Additional context**

These actions are responsible for a large portion of help desk tickets

hiemanshu:
This will create a new page that Superadmins can access through the user menu (on the top right, screenshot attached below).

The superadmin page will look similar to the main admin panel layout, with a sidebar on the left. It will start out with only 1 option in the sidebar, which will allow managing `People`.

In the people page, there will be a drop down to select the organization to manage, and will show a similar menu to the current organization's people page. In #1057 another option in the sidebar will be added to manage some organization settings. 

@bchrobot @politics-rewired/organizing thoughts?

hiemanshu:
<img width="305" alt="Screenshot 2022-07-25 at 4 56 38 PM" src="https://user-images.githubusercontent.com/367605/180769027-f135b01f-1649-40d9-8611-a516bea83d15.png">

bchrobot:
> In the people page, there will be a drop down to select the organization to manage, and will show a similar menu to the current organization's people page.

This ^^ should also allow viewing and managing other superadmins.

Otherwise, that all sounds good to me!

hiemanshu:
> > In the people page, there will be a drop down to select the organization to manage, and will show a similar menu to the current organization's people page.
> 
> This ^^ should also allow viewing and managing other superadmins.
> 
> Otherwise, that all sounds good to me!

So a new section (option in the sidebar) for this?

bchrobot:
> > > In the people page, there will be a drop down to select the organization to manage, and will show a similar menu to the current organization's people page.
> > 
> > 
> > This ^^ should also allow viewing and managing other superadmins.
> > Otherwise, that all sounds good to me!
> 
> So a new section (option in the sidebar) for this?

Either a new section or a "Superadmins" option at the top of the organization-selection dropdown. I think I lean more towards the latter. There could also be a "No Organization Memberships" dropdown option for texters who have not joined an organization for whatever reason.

ninabaldwin:
[...]

PR #1334, opened 2022-08-03
feat: add a superadmin page, add people management

hiemanshu:
## Description

Add a superadmin page where a superadmin can manage people from all orgs.

## Motivation and Context

Fixes #783 

## How Has This Been Tested?

Tested Locally

## Screenshots (if appropriate):

Screenshot 2022-08-03 at 2 23 31 PM
Screenshot 2022-08-03 at 2 23 39 PM
Screenshot 2022-08-03 at 2 23 47 PM

## Documentation Changes
Documentation will be required for this page. 

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have included updates for the documentation accordingly.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202732920779586

PR #1347, opened 2022-08-09
feat(superadmin): allow managing other superadmins

hiemanshu:
## Description

Allow managing other superadmins in the superadmin page

## Motivation and Context

Part 2 of #783 

## How Has This Been Tested?

Tested locally

Screenshot 2022-08-09 at 7 40 58 PM
Screenshot 2022-08-09 at 7 41 14 PM
Screenshot 2022-08-09 at 7 41 17 PM
Screenshot 2022-08-09 at 7 41 43 PM
Screenshot 2022-08-09 at 7 41 48 PM
Screenshot 2022-08-09 at 7 41 52 PM

## Checklist:


hiemanshu:
@politics-rewired/organizing ping for copy / UI feedback!

libs/spoke-codegen/src/graphql/super-admin-people.graphql
```diff
@@ -1,6 +1,41 @@
...
+    isSuperadmin
+    isSuspended
+  }
+}
+
+query GetPeople {
+  people {
+    ...Users
+  }
+}
```
ajohn25, line 31: Question: Any reason for creating the Users fragment and not a Superadmins fragment for these similar queries?

src/components/AddSuperAdminDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+              </Button>
+            </DialogActions>
+          </div>
+        )}
+      </Dialog>
+      <Dialog open={confirmDialogOpen} onClose={handleConfirmDialogClose}>
+        <DialogTitle>Confirm add superadmin?</DialogTitle>
+        <DialogContent>
+          <DialogContentText>
+            Are you sure you want to make this user ({userEmail}) a superadmin?
```
ajohn25, line 93: We think we should try to be extra extra careful around picking the right person throughout this page!
```suggestion
            Are you sure you want to make this user ({user.displayName} [{user.email}]) a superadmin?
```

src/components/Navigation.tsx
```diff
@@ -25,7 +25,7 @@ export interface NavigationSection {
   path: string;
   role?: string;
   badge?: { count: number };
-  url?: string;
+  url: string;
```
ajohn25, line 28: Question: Was this just a temporary change from the last PR in #1334 that's reverted now that a URL is always passed or more to it?

src/containers/SuperAdminSuperAdmin.tsx
```diff
@@ -0,0 +1,177 @@
...
+        onSubmit={handleAddSuperAdmin}
+      />
+      <Dialog
+        open={confirmRemoveSuperAdminDialog}
+        onClose={handleConfirmRemoveSuperAdminDialogClose}
+      >
+        <DialogTitle>Confirm remove superadmin</DialogTitle>
+        <DialogContent>
+          <DialogContentText>
+            Are you sure you want to remove the superadmin?
```
ajohn25, line 158: Suggestion to mimic the Add Superadmin Dialog if possible
```suggestion
            Are you sure you want to remove the superadmin ({user.displayName} [{user.email}])?
```

src/server/api/schema.js
```diff
@@ -3541,7 +3555,13 @@ const rootResolvers = {
       { organizationId, cursor, campaignsFilter, role },
       { user }
     ) => {
-      await accessRequired(user, organizationId, "SUPERVOLUNTEER");
+      if (organizationId) {
+        await accessRequired(user, organizationId, "SUPERVOLUNTEER");
+      } else if (!user.is_superadmin) {
+        throw new ForbiddenError(
```
ajohn25, line 3561: Suggestion: Can this be simplified to
```suggestion
      if (organizationId) {
        await accessRequired(user, organizationId, "SUPERVOLUNTEER", true);
      } else {
        throw new ForbiddenError(
```

Also outside the scope of this PR but I wonder if we should create an issue to change `accessRequired` to default to `allowSuperadmin = true` There's a note from way back in [`EXPLANATION-development-guidelines.md`](https://github.com/politics-rewired/Spoke/blob/main/docs/EXPLANATION-development-guidelines.md?plain=1#L142) about the decision but I'm not sure that in the way that superadmin has evolved, we would ever call this without also intending to let superadmins perform whatever action?

src/server/api/schema.js
```diff
@@ -3810,6 +3830,14 @@ const rootResolvers = {
         prevCampaignId,
         nextCampaignId
       };
+    },
+    superadmins: async (_root, _options, { user }) => {
+      if (user.is_superadmin) {
```
ajohn25, line 3835: Question: Should this be available to supervols and higher instead? Ex. On the regular People page, all users with access can view who's a superadmin. The People page of course doesn't currently refer to this field, but could imagine future references to it which are similar to that.

src/server/api/user.js
```diff
@@ -165,7 +167,8 @@ export const resolvers = {
       "isSuspended"
     ]),
     isSuperadmin: (userRecord, _, { user: authUser }) => {
-      if (userRecord.id !== authUser.id) throw new ForbiddenError();
+      if (userRecord.id !== authUser.id && !authUser.is_superadmin)
```
ajohn25, line 170: Question: See previous comment. Possibly may want users with lower permissions to have access to who's a superadmin or not.

ajohn25 (changes requested):
Nice! Some small suggestions for consistency below

libs/spoke-codegen/src/graphql/super-admin-people.graphql
hiemanshu: No specific reason, just been following the convention of `Data` fragment from `DataList` types. Will change this to `SuperAdmins`

src/components/Navigation.tsx
hiemanshu: Correct. The URL is a required parameter to actually jump between the different sections.

src/components/AddSuperAdminDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <DialogContent>
+              <DialogContentText>
+                Select user to promote to superadmin
+              </DialogContentText>
+              <FormControl>
+                <InputLabel id="add-superadmin-label">Pick User</InputLabel>
+                <Select
+                  style={{ width: 500 }}
+                  labelId="add-superadmin-label"
+                  value={userEmail}
```
bchrobot, line 69: Required: consistently pass value to prevent component from switching back and forth between controlled and uncontrolled

Either

```suggestion
                  value={userEmail ?? ""}
```

or changing state type to

```ts
const [userEmail, setUserEmail] = useState<string>("");
```

and the onChange handler to:

```ts
setUserEmail(event.target.value ?? "");
```

src/containers/SuperAdminSuperAdmin.tsx
```diff
@@ -0,0 +1,177 @@
...
+          }}
+        >
+          <DeleteForeverIcon />
+        </IconButton>
+      )
+    }
+  ];
+  return loading ? (
+    <div>Loading...</div>
+  ) : (
```
bchrobot, line 124: Suggestion 1: early return instead of ternary operator

```suggestion
  if (loading) {
    return <div>Loading...</div>;
  }
  
  return (
```

Suggestion 2: Return [`<Skeleton />`](https://v4.mui.com/components/skeleton/)

bchrobot (changes requested):

src/server/api/schema.js
hiemanshu: I think if we need this for a specific use case in the future, we can evaluate it then, currently superadmins are only accessible to other superadmins, and themselves. (below comment). I don't see a specific reason to change this behavior, and would prefer to keep permissions closed down as default, rather than opening up for a future possibility. 

@bchrobot thoughts?

bchrobot (approved):

src/server/api/schema.js
bchrobot: Sorry, I thought I had responded to this a week ago. I agree with @hiemanshu that keeping things closed down by default and opening up later if/when we have a use case

ajohn25: Sounds good 👍🏾

ajohn25 (approved):

Issue #1213, opened 2022-05-17
Simplify auto-assignment views


bchrobot:
Autoassignment relies on 8 Postgres views and makes 50-100+ line queries against them.

A design doc showing relationships between:

- React components -> GraphQL resolvers
- GraphQL resolvers -> lib methods
- Worker/other internals -> lib methods
- lib methods -> Postgres views
- Postgres views -> Postgres views

would be helpful and also likely illuminate opportunities for simplification such as `assignable_campaign_within_sending_time` suggested [here](https://github.com/politics-rewired/Spoke/pull/1211#discussion_r874064853).

PR #1350, opened 2022-08-12
feat(autosend): add basic autosend page

### LINKED ISSUE #1245: Autosending has long page load time
**Describe the bug**

The autosending page takes a long time to load, especially if there are many eligible campaigns.

**To Reproduce**

1.  Create ~10 autosend-eligible campaigns
2.  Go to Autosending page
3.  Experience long load time

**Expected behavior**

The page should load quickly.

**Screenshots**

N/A

**Additional context**

Probably need to tweak caching settings. Pre-fetching data in earlier resolvers may also help.

### LINKED ISSUE #1275: Multiple autosending campaigns can show as "sending"
**Describe the bug**

Only one campaign should show as "sending" at a time, based on the contact queue prioritization by campaign ID [here](https://github.com/politics-rewired/Spoke/blob/1d7c0b36f70b6eeffc56b469785f2fdd9b4ad220/src/server/tasks/queue-autosend-initials.ts#L53). However, "sending" vs "up next" is determined by whether _any_ messages have been sent on the campaign (see [here](https://github.com/politics-rewired/Spoke/blob/1d7c0b36f70b6eeffc56b469785f2fdd9b4ad220/src/containers/AdminAutosending/components/AutosendingTargetRow.tsx#L48-L55)), not contact queue prioritization.

**To Reproduce**

1.  Create and start two campaigns
2.  Manually send 1 message on each campaign
3.  Queue both campaigns for autosending
4.  See both campaigns with status "sending"

**Expected behavior**

Only one campaign should show as sending at a time, and it should match the contact prioritization.

Something like the following:

```sql
select
  (case
    when autosend_status = 'sending' and id <> (select min(id) from campaign where autosend_status = 'sending' and is_started and not is_archived) then 'up next'
    else autosend_status
  end) as autosend_status
from campaign
where
  -- ...
```

**Screenshots**

N/A

**Additional context**

N/A

ajohn25:
## Description
This adds a Basic mode for the autosending page for faster loading.
## Motivation and Context
Closes #1245. Closes #1275.

## How Has This Been Tested?
This has been tested locally.

<table>
<tr>
<th>Basic</th>
<th>Detailed</th>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist:


migrations/20220711050742_add_autosend_view.js
```diff
@@ -0,0 +1,115 @@
...
+          select id, title, organization_id, limit_assignment_to_teams, autosend_status
+          from campaign
+          where is_started = true
+            and is_archived = false
+            and is_autoassign_enabled = true
+        );
+      `
+    )
+    .then(() => {
+      return knex.schema.raw(`
```
bchrobot, line 85: Question: do these need to be broken up into two batches?

migrations/20220711050742_add_autosend_view.js
```diff
@@ -0,0 +1,115 @@
...
+            select 1
+            from assignable_needs_message
+            where campaign_id = sendable_campaigns.id
+          )
+          and not exists (
+            select 1
+            from campaign
+            where campaign.id = sendable_campaigns.id
+              and now() > date_trunc('day', (due_by + interval '24 hours') at time zone campaign.timezone)
+          )
```
bchrobot, line 58: Question: is there a reason not to do this as a join? It feels a little odd to use exists to check on a condition on the original row itself

src/containers/AdminAutosending/index.tsx
```diff
@@ -19,28 +19,47 @@ import {
   usePauseAutosendingMutation,
   useStartAutosendingMutation
 } from "@spoke/spoke-codegen";
+import { css, StyleSheet } from "aphrodite";
 import { sortBy } from "lodash";
 import React, { useCallback, useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 import { BooleanParam, useQueryParam, withDefault } from "use-query-params";
 
+import { AutosendingControlsMode } from "../../../libs/spoke-codegen/src/generated";
```
bchrobot, line 28: Required: prefer `@spoke/spoke-codegen`

src/containers/AdminAutosending/index.tsx
```diff
@@ -19,28 +19,47 @@ import {
...
+    AutosendingControlsMode.Basic;
+
   const [isStarted = true, setIsStarted] = useQueryParam(
     "isStarted",
     withDefault(BooleanParam, true)
   );
+  const [isBasic = defaultBasic, setIsBasic] = useQueryParam(
+    "isBasic",
+    withDefault(BooleanParam, defaultBasic)
+  );
```
bchrobot, line 54: Question: why is `isBasic = defaultBasic` required when using `withDefault()`?

src/containers/AdminAutosending/index.tsx
```diff
@@ -80,6 +99,27 @@ const AdminAutosending: React.FC = () => {
...
+      const isChangeToBasic =
+        event.target.value === AutosendingControlsMode.Basic;
+      if (isChangeToBasic) setIsBasic(isChangeToBasic);
+      else
+        refetchCampaigns({
+          organizationId,
+          isStarted,
+          isBasic: false
+        }).then(() => setIsBasic(isChangeToBasic));
+    },
```
bchrobot, line 119: Question: is the manual refetch necessary? `isBasic` is passed to `useCampaignsEligibleForAutosendingQuery()` variables and a change to variables should trigger a refetch. Maybe a problem with the Apollo fetch policy for the query if that's not happening?

src/containers/AdminAutosending/index.tsx
```diff
@@ -133,51 +172,98 @@ const AdminAutosending: React.FC = () => {
...
+                onChange={handleChangeStartedFilter}
+              >
+                <MenuItem value="started">Started</MenuItem>
+                <MenuItem value="unstarted">Unstarted</MenuItem>
+              </Select>
+            </FormControl>
+            <FormControl>
+              <InputLabel id="is-started-select-label">
+                Campaign Status
+              </InputLabel>
```
bchrobot, line 194: Required: update label

```suggestion
              <InputLabel id="autosend-controls-mode-select-label">
                Display Mode
              </InputLabel>
```

src/containers/Settings/components/AutosendingSettingsCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+        controlsMode
+      }
+    });
+  };
+
+  return (
+    <Card style={style}>
+      <CardHeader title="Autosending Settings" disableTypography />
+      <CardContent>
+        {errorMsg && <p>Error: {errorMsg}</p>}
```
bchrobot, line 61: Required: prefer [`<Alert />`](https://v4.mui.com/components/alert/)

bchrobot (changes requested):

migrations/20220711050742_add_autosend_view.js
ajohn25: Does not actually, I'll combine them

migrations/20220711050742_add_autosend_view.js
ajohn25: This is copied from the same due date restriction in [`assignable_campaigns_with_needs_message`](https://github.com/politics-rewired/Spoke/pull/1350/files#diff-e3460a39ce585fb85dbbac59110b9f6ccf3ca2f3aa0d3cf1f8a1212c5408429cR25) I think it might make sense to leave it this way to keep it clear they're connected and leave a note for the rewrite as part of #1213?

src/containers/AdminAutosending/index.tsx
ajohn25: I don't think it actually is. In the existing query param setup for `isStarted`, I saw we were explicitly setting `isStarted = true` so it seemed like we wanted to be extra careful around setting the default. But I don't see any difference in behavior when removing that so seems fine to me to take out.

src/containers/AdminAutosending/components/AutosendingBasicTargetRow.tsx
```diff
@@ -0,0 +1,103 @@
...
+import MoreIcon from "@material-ui/icons/ArrowForward";
+import PauseIcon from "@material-ui/icons/Pause";
+import PlayIcon from "@material-ui/icons/PlayArrow";
+import { AutosendingTargetFragment } from "@spoke/spoke-codegen";
+import React from "react";
+import { Link } from "react-router-dom";
+
+const useStyles = makeStyles({
+  unstarted: {
+    backgroundColor: grey[200]
```
hiemanshu, line 18: Required: Use theme palette for these colors. If it doesn't exist in theme, create them!

src/containers/AdminAutosending/components/AutosendingBasicUnstartedTargetRow.tsx
```diff
@@ -0,0 +1,50 @@
...
+import TableCell from "@material-ui/core/TableCell";
+import TableRow from "@material-ui/core/TableRow";
+import MoreIcon from "@material-ui/icons/ArrowForward";
+import { AutosendingTargetFragment } from "@spoke/spoke-codegen";
+import React from "react";
+import { Link } from "react-router-dom";
+
+const useStyles = makeStyles({
+  unstarted: {
+    backgroundColor: grey[300]
```
hiemanshu, line 13: Required: Use theme palette

src/containers/AdminAutosending/index.tsx
```diff
@@ -19,28 +19,47 @@ import {
   usePauseAutosendingMutation,
   useStartAutosendingMutation
 } from "@spoke/spoke-codegen";
+import { css, StyleSheet } from "aphrodite";
```
hiemanshu, line 22: Suggestion: Prefer using `makeStyles`

hiemanshu (changes requested):
Looks good to me! Some small changes around css / theme palette management.

src/containers/AdminAutosending/index.tsx
```diff
@@ -15,26 +16,25 @@ import TableHead from "@material-ui/core/TableHead";
...
 import Alert from "@material-ui/lab/Alert";
 import {
+  AutosendingControlsMode,
   useCampaignsEligibleForAutosendingQuery,
   usePauseAutosendingMutation,
   useStartAutosendingMutation
 } from "@spoke/spoke-codegen";
-import { css, StyleSheet } from "aphrodite";
-import { sortBy } from "lodash";
+import { isNil, sortBy } from "lodash";
```
bchrobot, line 24: Required: import specific functions to [support tree shaking](https://www.azavea.com/blog/2019/03/07/lessons-on-tree-shaking-lodash/) during compilation

```suggestion
import isNil from "lodash/isNil";
import sortBy from "lodash/sortBy";
```

hiemanshu (approved):

PR #1351, opened 2022-08-12
fix(message-review): display accurate error description

bchrobot:
## Description

When possible, display the correct human-friendly error message for a message.

## Motivation and Context

We currently report all errored messages "rejected by carrier." This was a reasonable assumption when Spoke did not track error codes and had no knowledge about why a message errored. We now track that and already have a mapping of error codes to human-friendly descriptions.

## How Has This Been Tested?

This has been tested locally.

Screen Shot 2022-08-12 at 3 07 30 PM

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/lib/telco-error-codes.tsx
+++ b/src/lib/telco-error-codes.tsx
@@ -1,0 +1,39 @@
+export const errorCodeDescriptions: Record<string, string> = {
+  "40001": "Invalid destination number",
+  "40002": "Blocked as spam",
+  "40003": "Blocked as spam",
+  "40004": "Unknown",
+  "40005": "Expired",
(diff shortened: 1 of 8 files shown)


ajohn25 (approved):

src/lib/telco-error-codes.tsx
```diff
@@ -0,0 +1,39 @@
+export const errorCodeDescriptions: Record<string, string> = {
```
hiemanshu, line 1: Required: File should be `.ts` and not `.tsx`

hiemanshu (approved):
LGTM! Just one small required change.

PR #1357, opened 2022-08-21
fix(campaign builder): fix deleting interaction steps

ajohn25:
## Description
This fixes the deletion of interaction steps when using the superadmin approval feature.

## Motivation and Context
When a non superadmin edits the script on a started campaign when using superadmin approval, the campaign is unstarted. Messaged contacts on an unstarted campaign were previously (mostly) impossible, and so we used the `is_started` status of the campaign to check whether interaction steps could be safely deleted.

However, in this case with superadmin approval, a campaign can be unstarted even if contacts are messaged. If survey responses have been applied, deletion of a record in `interaction_step` violates the foreign key constraint on `question_response`. This blocks any script edits from taking effect, after deleting an interaction step has been attempted.

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/__test__/testbed-preparation/core.ts
+++ b/__test__/testbed-preparation/core.ts
@@ -1,28 +1,31 @@
 import faker from "faker";
 import AuthHasher from "passport-local-authenticate";
-import { PoolClient } from "pg";
+import type { PoolClient } from "pg";
 
-import { Assignment } from "../../src/api/assignment";
-import { Campaign } from "../../src/api/campaign";
-import { CampaignContact } from "../../src/api/campaign-contact";
-import { InteractionStep } from "../../src/api/interaction-step";
-import { Message } from "../../src/api/message";
-import { Organization } from "../../src/api/organization";
+import type { Assignment } from "../../src/api/assignment";
+import type { Campaign } from "../../src/api/campaign";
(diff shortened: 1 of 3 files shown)


bchrobot (changes requested):
Required: check against the referencing tables explicitly

Presence of messages is a good proxy for the presence of question responses but `all_question_response` rows are the ones that will cause a foreign key exception to be raised. An admin _could_ mark a question response from Message Review before a campaign is started and end up in the same situation without any messages being sent. `interaction_step` is also referenced by `all_external_sync_question_response_configuration` without any `on delete` behavior on the foreign key.

One option might be to [`.clone()`](https://knexjs.org/guide/query-builder.html#clone) the query, `try {}` it with `.del()`, and then `.update()` if the caught error is a Postgres foreign key constraint exception.

hiemanshu (approved):
~~Overall LGTM. I think this should throw out an error to the user to show why it's not being deleted, otherwise the UI wouldn't change, and it would seem like a broken UI. @politics-rewired/organizing thoughts?~~


nvm, ignore me. I read the code wrong.

bchrobot (approved):

hiemanshu (approved):

PR #1359, opened 2022-08-23
feat: add service worker to cache client bundles

### LINKED ISSUE #1343: Add a serviceWorker
I believe this article has some good basics: https://css-tricks.com/add-a-service-worker-to-your-site/

This library seems to be useful for managing cache status: https://developer.chrome.com/docs/workbox/

Our service worker should cache all CSS and JS assets. These are content addressed and contain a hash, so there's no harm to doing so :)

Our service worker should additionally cache `index.html`, but should pay attention to the `x-spoke-version` header on all requests that it intercepts (including GraphQL requests). If it finds that the `x-spoke-version` header has changed, it should purge it's `index.html` cache.

Once this is implemented, in combination with https://github.com/politics-rewired/Spoke/issues/1342, any user who has successfully visited the page before and not manually uninstalled the service worker will not get a 503.

hiemanshu:
## Description

Adds a service worker using workbox to cache the javascript bundles

## Motivation and Context

Fixes #1343

## How Has This Been Tested?

Tested locally, will require some bits of extensive testing on staging env

## Checklist:


src/client/index.tsx
```diff
@@ -16,3 +18,11 @@ const main = (settings: InstanceSettings) => {
...
 
 request.get("/settings/instance").then(({ body: settings }) => main(settings));
+
+// Register service worker
+serviceWorkerRegistration.register();
+
+// If you want to start measuring performance in your app, pass a function
+// to log results (for example: reportWebVitals(console.log))
+// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
+reportWebVitals();
```
bchrobot, line 28: Question: are we using web vitals for anything?

src/client/service-worker.ts
```diff
@@ -0,0 +1,84 @@
...
+// Their URLs are injected into the manifest variable below.
+// This variable must be present somewhere in your service worker file,
+// even if you decide not to use precaching. See https://cra.link/PWA
+precacheAndRoute(self.__WB_MANIFEST);
+
+// Set up App Shell-style routing, so that all navigation requests
+// are fulfilled with your index.html shell. Learn more at
+// https://developers.google.com/web/fundamentals/architecture/app-shell
+const fileExtensionRegexp = /\/[^/?]+\.[^/]+$/;
+registerRoute(
```
bchrobot, line 31: Question: does this need a second argument like `new StaleWhileRevalidate()` below? Where is it specified that the same content served by the API on initial page load (e.g. `index.html`) is to be served (and cached)? `__WB_MANIFEST`?

src/client/serviceWorkerRegistration.ts
```diff
@@ -0,0 +1,146 @@
...
+          }
+        };
+      };
+    })
+    .catch((error) => {
+      console.error("Error during service worker registration:", error);
+    });
+}
+
+function checkValidServiceWorker(swUrl: string, config?: Config) {
```
bchrobot, line 72: Suggestion: rename to `checkAndRegisterValidServiceWorker()`

```suggestion
function checkAndRegisterValidServiceWorker(swUrl: string, config?: Config) {
```

or move the call to `registerValidSW()` outside to just before `navigator.serviceWorker.ready()`

webpack.config.ts
```diff
@@ -41,6 +42,18 @@ if (config.isProduction) {
       publicPath: ""
     })
   );
+
+  plugins.push(
+    new InjectManifest({
+      swSrc: "./client/service-worker",
+      dontCacheBustURLsMatching: /\.[0-9a-f]{8}\./,
+      exclude: [/\.map$/, /asset-manifest\.json$/, /LICENSE/],
```
bchrobot, line 50: Question: should this also exclude server-side files? Or instead use `include` for client assets?

webpack.config.ts
```diff
@@ -41,6 +42,18 @@ if (config.isProduction) {
...
+      // Bump up the default maximum size (2mb) that's precached,
+      // to make lazy-loading failure scenarios less likely.
+      // See https://github.com/cra-template/pwa/issues/13#issuecomment-722667270
+      maximumFileSizeToCacheInBytes: 5 * 1024 * 1024
```
bchrobot, line 54: Suggestion: make this dependent on `NODE_ENV`. Our production bundle size is ~4mb but development bundle is ~30mb, I believe.

bchrobot (commented):
Mostly just questions at this point!

ajohn25 (approved):
Mostly outside my ability to give a good review so also suggesting to add @ben-pr-p for 2nd 👀 but no flags from me!

src/client/serviceWorkerRegistration.ts
```diff
@@ -0,0 +1,146 @@
...
+    })
+    .catch(() => {
+      console.log(
+        "No internet connection found. App is running in offline mode."
+      );
+    });
+}
+
+export function register(config?: Config) {
+  if (process.env.NODE_ENV === "production" && "serviceWorker" in navigator) {
```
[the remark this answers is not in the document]
hiemanshu: Yup.

Dockerfile
```diff
@@ -66,6 +65,7 @@ ENV NODE_ENV="production" \
 COPY package.json knexfile.env.js ./
 COPY migrations ./migrations
 COPY seeds ./seeds
+RUN cp ./build/client/assets/service-worker.js ./build/client/
```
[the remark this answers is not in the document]
hiemanshu: `service-worker.js` is built using Workbox's `InjectManifest` it ends up in assets because that's the defined output folder for the webpack config. 

```  
output: {
  filename: config.isProduction ? "[name].[chunkhash].js" : "[name].js",
  path: path.resolve(config.isProduction ? config.ASSETS_DIR : __dirname),
  publicPath: "/assets/"
}
```

Because of this config, it ends up `assets/` with all the other files, and maps. It is being moved out of there to `build/client` because that is defined as the `PUBLIC_DIR` (https://github.com/politics-rewired/Spoke/blob/feature/service-worker/Dockerfile#L57) and served by express directly (https://github.com/politics-rewired/Spoke/blob/feature/service-worker/src/server/app.ts#L89). Making any changes to this setup was causing a cascading of failures, like the bundle not being found by the frontend, etc. Trying to access the file through `/assets/service-worker.js` would return the bundle file, and not the service worker. This is the easiest way that work. I expect this to be refactored when we do CDN setup, so that paths can be more appropriate.

src/client/index.tsx
hiemanshu: Not right now.

Dockerfile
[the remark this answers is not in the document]
bchrobot: > Trying to access the file through /assets/service-worker.js would return the bundle file, and not the service worker.

Ah, this was the key piece I didn't understand. That's very annoying.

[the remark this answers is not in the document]
bchrobot: Question: how is invalidation of the service worker itself managed with no build hash to go off of?

webpack.config.ts
hiemanshu: We don't want to setup the service worker for development (annoying cache issues), which is why the plugin is only enabled in `production` mode, and 5MB is a safe size for now.

Dockerfile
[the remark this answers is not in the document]
hiemanshu: The browsers handle it automatically! https://web.dev/service-worker-lifecycle/#updates

> Your service worker is considered updated if it's byte-different to the one the browser already has. (We're extending this to include imported scripts/modules too.)

bchrobot (approved):

hiemanshu:
screenshot_2022-09-14_at_1 30 56_pm

Looks like caches are working as expected! Going to go ahead and merge this PR, and we can make future enhancements on these as necessary.

PR #1362, opened 2022-08-24
feat: show a snackbar on mulitple graphql server errors

### LINKED ISSUE #1342: Extend `apollo-client-singleton` to gracefully handle non JSON responses
We can hook into requests that are about to be passed to Apollo [here](https://github.com/politics-rewired/Spoke/blob/96f88f378e86ad1345723bf20f332b5d779c5243/src/network/apollo-client-singleton.ts).

Right now, if the server returns a 503 or other unhandled error code or otherwise a non JSON response, we attempt to parse that HTML as JSON, and return to the caller of the mutation an error that says something like "Unable to parse <! as JSON".

Instead, we should intercept that condition based on the `content-type` header and do 2 things:
-return a well formed JSON object that looks like:
```typescript
{ loading: false, data: undefined, errors: ["We were unable to properly reach the server, or the request might have timed out. Please try again momentarily"] }
```
- Emit a message on the `eventBus.emit` called something like `EventTypes.ServerUnreachable`

Finally, we can create a component which is close to a clone of [VersionNotifier](https://github.com/politics-rewired/Spoke/blob/00a408a5aee44c5b655cc4b0c0ec2a33edde7b8d/src/client/VersionNotifier.tsx) that tracks how many `ServerUnreachable`s we get. After maybe 2 or 3, we should pop up maybe a [Snackbar](https://mui.com/material-ui/react-snackbar/) informing the user that we're having consistent issues communicating with the server, and they should wait a minute and refresh.

hiemanshu:
## Description

Emit errors on the eventbus when a graphql server error is encountered, and show a snackbar 

## Motivation and Context

Fixes #1342 

## How Has This Been Tested?

Tested locally with a few scenarios. Will require some extra testing on staging. 

## Checklist:

### DIFF
--- a/src/client/ErrorHandler.tsx
+++ b/src/client/ErrorHandler.tsx
@@ -1,0 +1,56 @@
+import Snackbar from "@material-ui/core/Snackbar";
+import Alert from "@material-ui/lab/Alert";
+import React, { useEffect, useRef, useState } from "react";
+
+import { eventBus, EventTypes } from "./events";
+
+const ErrorHandler: React.FC = () => {
+  const errorCount = useRef(0);
+  const [open, setOpen] = useState<boolean>(false);
+
+  const handleNewEvent = () => {
+    errorCount.current += 1;
+
+    // We have our first event triggered
+    if (errorCount.current === 1) {
+      // Set timeout to expire count after 30 seconds
+      // so a few events over a long period of time
+      // don't set off the snackbar
+      setTimeout(() => {
+        errorCount.current = 0;
+      }, 30 * 1000);
+    }
+
+    // Show the snackbar, and reset the count to 0
+    if (errorCount.current > 3) {
+      setOpen(true);
+      errorCount.current = 0;
+    }
+  };
+
+  const handleClose = () => setOpen(false);
+
+  useEffect(() => {
+    eventBus.on(EventTypes.GraphQLServerError, handleNewEvent);
+  }, []);
+
+  return (
+    <Snackbar
+      open={open}
+      autoHideDuration={6000}
+      onClose={handleClose}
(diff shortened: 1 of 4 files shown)


ajohn25 (approved):

src/client/ErrorHandler.tsx
```diff
@@ -0,0 +1,56 @@
...
+      setOpen(true);
+      errorCount.current = 0;
+    }
+  };
+
+  const handleClose = () => setOpen(false);
+
+  useEffect(() => {
+    eventBus.on(EventTypes.GraphQLServerError, handleNewEvent);
+  }, []);
```
bchrobot, line 35: Required: this should return a [clean-up function](https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect) to remove the listener when the component unmounts. See VersionNotifier:

https://github.com/politics-rewired/Spoke/blob/121f9234ad5f3ed1a8712548840b07b7e4beba6e/src/client/VersionNotifier.tsx#L22-L30

src/client/ErrorHandler.tsx
```diff
@@ -0,0 +1,56 @@
...
+    errorCount.current += 1;
+
+    // We have our first event triggered
+    if (errorCount.current === 1) {
+      // Set timeout to expire count after 30 seconds
+      // so a few events over a long period of time
+      // don't set off the snackbar
+      setTimeout(() => {
+        errorCount.current = 0;
+      }, 30 * 1000);
```
bchrobot, line 21: Required: the timer ID should be stored in a ref and cleared (if set) when the component unmounts. This can probably be done in the same clean up function as the event bus listener (see comment below).

bchrobot (changes requested):

bchrobot (approved):

PR #1365, opened 2022-08-26
fix: import server-defined type

bchrobot:
## Description

Import `DeactivateMode` from server-defined type rather than client-defined type.

## Motivation and Context

`@spoke/spoke-codegen` is meant for use in the client application only and will fail to import in production.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:


hiemanshu (approved):
Interesting, Didn't know this could break!

hiemanshu:
Looks like type imports are fine, because TS will remove that out when compiling, but enums are referred to, so causes an issue. LGTM!

ajohn25 (approved):

PR #1366, opened 2022-08-26
style: import codegen using alias rather than path

bchrobot:
## Description

Require importing codegen using alias rather than relative path.

## Motivation and Context

Codegen is an internal library and should be used like one.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -100,6 +100,10 @@ module.exports = {
         ],
         patterns: [
           {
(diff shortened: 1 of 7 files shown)


hiemanshu (approved):
LGTM!

ajohn25 (approved):
Thought of doing this after the last time I  used the wrong one - beat me to it 🎉

PR #1367, opened 2022-08-26
style: require type imports

bchrobot:
## Description

Make type vs value imports explicit.

## Motivation and Context

This can help reasoning about imports and makes the TS compiler's job a little easier.

All changes beyond `.eslintrc.js` are from `yarn lint:fix`.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:


ajohn25 (approved):

hiemanshu (approved):
LGTM! Just a couple of things I think we should note down to figure out for later.

1. Types should be imported from spoke codegen, so we don't define schema AND export the type as well. 
2. How to deal with enums that are common in the frontend/backend. So they can be imported by being defined only once, rather than having a graphql version and a backend version.

PR #1395, opened 2022-08-29
feat: add opt outs page, and allow opt in

### LINKED ISSUE #833: Add option to opt out entire campaign
In order to opt out a list of contacts, admins must:

1. Load a dummy campaign with a list of contacts
1. Email support and request that contacts present on that campaign be opted out

There is no current method for admins to opt out contacts en masse.

"Opt out whole campaign" and "Opt in whole campaign" should be an option during campaign creation. A warning message should appear before executing the action to confirm understanding of implications.

Opt out lists could be managed through a new individual page view rather than the Campaigns page.

**Additional context**
When organizations join, they sometimes bring six-figure lists of contacts who were opted out on other platforms. If requests to opt out lists of that size would likely fail, other solutions should be discussed.

hiemanshu:
## Description

Add opt outs page to view opt outs by campaign, and import opt outs and opt ins. 

## Motivation and Context

Fixes #833 

## How Has This Been Tested?
Tested locally

Screenshot 2022-08-29 at 8 54 35 PM
Screenshot 2022-08-29 at 8 54 41 PM
Screenshot 2022-08-29 at 8 54 44 PM
Screenshot 2022-08-29 at 8 54 48 PM
Screenshot 2022-08-29 at 8 54 52 PM

## Checklist:

### DIFF
--- a/libs/spoke-codegen/src/graphql/opt-outs.graphql
+++ b/libs/spoke-codegen/src/graphql/opt-outs.graphql
@@ -1,4 +1,4 @@
-query GetOptOutsByCampaign($organizationId: String!) {
+query GetOptOutsByCampaignForOrganization($organizationId: String!) {
   optOuts(organizationId: $organizationId) {
     id
     title
--- a/schema-dump.sql
+++ b/schema-dump.sql
@@ -2833,7 +2833,7 @@ ALTER SEQUENCE public.notification_id_seq OWNED BY public.notification.id;
 CREATE TABLE public.opt_out (
     id integer NOT NULL,
     cell text NOT NULL,
-    assignment_id integer NOT NULL,
+    assignment_id integer,
     organization_id integer NOT NULL,
     reason_code text DEFAULT ''::text NOT NULL,
     created_at timestamp with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
--- a/src/containers/AdminOptOutList/index.tsx
+++ b/src/containers/AdminOptOutList/index.tsx
@@ -10,7 +10,7 @@ import { DataGridPro } from "@mui/x-data-grid-pro";
 import {
   useBulkOptInMutation,
   useBulkOptOutMutation,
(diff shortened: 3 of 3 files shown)


hiemanshu:
@politics-rewired/organizing feedback if any as usual!

@bchrobot do you think the bulk opt in or bulk opt out should be a job processed by graphile? 

PR'ing it a little early to get as much feedback as possible before spending more time rounding things up.

libs/spoke-codegen/src/graphql/opt-outs.graphql
```diff
@@ -0,0 +1,31 @@
+query GetOptOutsByCampaign($organizationId: String!) {
```
bchrobot, line 1: Suggestion: rename to `GetOptOutsByCampaignForOrganization`. Just "by campaign" along with the organization ID variable could be confusing

bchrobot (changes requested):
Required: update schema-dump.sql

Otherwise looks good!

hiemanshu:
@bchrobot please add a milestone as well!

bchrobot (approved):

src/api/schema.ts
```diff
@@ -229,6 +229,12 @@ const rootSchema = `
     campaignId: String!
   }
 
+  type OptOutByCampaign {
+    id: String!
+    title: String!
+    count: String!
```
ajohn25, line 235: Why is count a string here instead of int?

ajohn25 (approved):
Small question for my understanding but looks great!

src/api/schema.ts
hiemanshu: Just seeing this @ajohn25, Most data including IDs are passed back as String especially if they're only going to be displayed and not manipulated in any way. So going with `String!` in most scenarios to keep that convention up.

bchrobot: That convention of Int -> String should be limited to IDs, which GraphQL is opinionated about. We should try to match type as best we can for other properties.

PR #1411, opened 2022-09-11
feat: offline page

bchrobot:
## Description

This adds a default offline page to be returned when Spoke Rewired cannot be reached.

When Spoke Rewired _can_ be reached, the index is loaded from the server (where it is cached at startup) and the app bundle is served from the service worker cache.

## Motivation and Context

Spoke Rewired is not a progressive web application (PWA). Caching and returning the index page response only gets us part of the way there but still results in an ugly error page being displayed to the user when subsequent required API requests fail. Some of these requests, like `/settings/instance`, can safely be cached with a [`NetworkFirst` strategy](https://developer.chrome.com/docs/workbox/modules/workbox-strategies/#network-first-network-falling-back-to-cache). Others, such as requests to get the current user and application data should not be cached that way. Nor can they easily, as they are GraphQL requests a) using the same endpoint and b) made with `POST` method.

Quite a bit more work would need to be done to make Spoke Rewired a true PWA that could handle offline mode within the React app itself and have that be a good UX.

I could not get the muicss resources to load from the service worker cache so I just copied over the relevant classes. I believe this is a CORS issue as the font resource loads fine but has a few extra cors-related headers that muicss does not.

## How Has This Been Tested?

This has been tested locally:

```sh
NODE_ENV=production yarn dotenv -- yarn build
NODE_ENV=production yarn dotenv -- node build/src/server/index.js
```

Fullscreen_9_11_22__7_57_AM

## Documentation Changes

N/A

## Checklist:


src/server/app.ts
```diff
@@ -99,6 +100,18 @@ export const createApp = async () => {
...
   }
 
+  // Serve service worker from root for complete SW scope
+  app.get("/service-worker.js", (req, res) =>
+    res.sendFile("service-worker.js", { root: config.ASSETS_DIR })
+  );
+
+  // Serve offline page
+  app.get("/offline.html", (req, res) =>
+    res.sendFile("offline.html", {
```
hiemanshu, line 110: How is the file being copied over? I think webpack should copy over this file to the clients build directly. This will allow the clients build directory to later be uploaded to CDN more easily allowing static files to be available via CDN.

src/client/offline.html
```diff
@@ -0,0 +1,308 @@
+<!doctype html>
```
hiemanshu, line 1: Required: Static files should live in the `public/`

hiemanshu (approved):
Overall the changes LGTM! Just thinking about how we can do work to setup for upcoming features. And an offline page is a good addition! Thanks for all the work testing the service worker <3

src/client/offline.html
bchrobot: Spoke Rewired's `PUBLIC_DIR=./build/client` which only ever contains `/assets/` generated by webpack. So we don't really have a way of serving static files that aren't managed by webpack. I thought about making some tweaks to address that but it felt out of scope for the service worker PRs.

src/server/app.ts
bchrobot: It is copied to `build/` during `yarn build:server` (along with the entire `src/client/ directory 😔)

I was worried about scope creep making more webpack changes to handle a static html file but it turns out to be pretty straightforward. See latest commits.

hiemanshu: Yeah, it's a little out of scope of this PR, but I'd rather do things the right way with little scope creep, than make half way changes we'll want to fix later on, especially since we know we want to make those changes in the not so distant future.

src/client/offline.html
hiemanshu: As mentioned above, making some tweaks to get things setup correctly is good with me! Even with a little bit of scope creep.

The comment was about putting the source file in `public/` folders, and then copying it over to `/build/client` using webpack. Which I believe is now happening here: https://github.com/politics-rewired/Spoke/pull/1411/files#diff-ec2cabc8139e814d33682814c1ba15a1056d946ab70f740c9d42ada07af42050R41
This way all static files that we need to copy (in this case, a copy webpack plugin might be better than a html webpack plugin) are copied from `public/` into `PUBLIC_DIR`

bchrobot: Using a copy plugin over html sounds good to me as a future change!

(especially if it comes with a definitive asset management policy, reduction in environment variables, and path to CDN-hosting)

hiemanshu (approved):

PR #1360, opened 2022-08-23
fix: fix refresh integration button hidden

hiemanshu:
## Description

Set text overflow to clip, so it doesn't convert buttons to ellipsis. 

## Motivation and Context

Fixes #1283 

## How Has This Been Tested?
Tested locally

## Screenshots (if appropriate):

Screenshot 2022-08-23 at 9 22 16 PM
Screenshot 2022-08-23 at 9 22 37 PM

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have included updates for the documentation accordingly.

PR #1369, opened 2022-08-28
chore(mui): upgrade tables to v4

bchrobot:
## Description

Updates `<Table>` usage to MUI v4.

It also refactors `<InfiniteRelayList>` to remove intermediate components screwing up the table.

## Motivation and Context

Closes #1163 

## How Has This Been Tested?

This has been tested locally.

## Screenshots (if appropriate):

N/A

## Documentation Changes

N/A

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have included updates for the documentation accordingly.

PR #1413, opened 2022-09-11
fix: fix integration cell click handling

ajohn25:
## Description
This fixes click handling when clicking cells on the Integrations page.

## Motivation and Context
The MUI v4 [`<Table>`](https://v4.mui.com/api/table/) upgrade in #1360 prevented users from clicking on table cells to access the integration detail page, because the `onCellClick` event is no longer supported. The v4 [`<Table>`](https://v4.mui.com/api/table/) also doesn't support several other props we were using from v0, and are no longer required because of default behavior.

## How Has This Been Tested?
This has been tested locally.
## Checklist:

### DIFF
--- a/src/containers/AdminExternalSystems.tsx
+++ b/src/containers/AdminExternalSystems.tsx
@@ -191,6 +191,8 @@ class AdminExternalSystems extends Component<Props, State> {
       </Button>
     ];
 
+    const { organizationId } = this.props.match.params;
+
     return (
       <div>
         <FloatingActionButton
@@ -222,10 +224,9 @@ class AdminExternalSystems extends Component<Props, State> {
             </TableHead>
             <TableBody>
               {externalSystems.edges.map(({ node: system }) => {
-                const { organizationId } = this.props.match.params;
                 const detailURL = `/admin/${organizationId}/integrations/${system.id}`;
                 return (
-                  <TableRow key={system.id}>
+                  <TableRow key={system.id} hover>
                     <TableCell>
                       <Link to={detailURL}>{system.name}</Link>
                     </TableCell>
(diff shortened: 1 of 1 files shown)


src/containers/AdminExternalSystems.tsx
```diff
@@ -236,12 +227,22 @@ class AdminExternalSystems extends Component<Props, State> {
                 <TableCell>Actions</TableCell>
               </TableRow>
             </TableHead>
-            <TableBody displayRowCheckbox={false} showRowHover>
+            <TableBody>
               {externalSystems.edges.map(({ node: system }) => (
                 <TableRow key={system.id}>
```
bchrobot, line 232: Required: retain hover behavior

```suggestion
                <TableRow key={system.id} hover>
```

bchrobot (changes requested):

ajohn25:
> The approach taken in #1369 was to have the `onClick` handler on the `<TableRow>` and then `stopPropagation()` of the event in the button `onClick` handlers (see [`<TeamEditorList>` changes](https://github.com/politics-rewired/Spoke/pull/1369/files#diff-385ab67ec9a44510d8c74071a93dbda702a9ee9265b2a3e96056dae8fea9cd08)). I don't have strong feelings either way but we should try to be consistent.

Missed that existed! If it's just in [`<TeamEditorList>`](https://github.com/politics-rewired/Spoke/pull/1369/files#diff-385ab67ec9a44510d8c74071a93dbda702a9ee9265b2a3e96056dae8fea9cd08) where the `stopPropagation()` approach is being used currently, I think I prefer to make the change to cell `onClick` events there, which might feel more intuitive both for users (slightly) + developers. Lmk if you think differently!

hiemanshu (approved):
LGTM!

bchrobot (approved):

PR #1416, opened 2022-09-13
chore: deprioritize autosending

ajohn25:
## Description
This deprioritizes autosending jobs, so that jobs related to message sending are prioritized in the following order: autoassignment, autosending, handling delivery reports.

## Motivation and Context
On instances with high autosending throughput that also had autoassignment requests occur simulatenously, autoassignment requests could get "stuck in pending" because there were significantly more autosending jobs generated at a given time than autoassignment jobs, and workers never picked up the autoassignment jobs. When autoassignment has a higher priority, this should not occur so frequently.

## How Has This Been Tested?
This has been tested locally to ensure both autoassignment and autosending jobs are picked up by workers. We should also observe autosending performance on previously mentioned instances that have a high rate of autosending and autoassignment requests before a wide rollout. 

## Checklist:

### DIFF
--- a/src/server/tasks/queue-autosend-initials.ts
+++ b/src/server/tasks/queue-autosend-initials.ts
@@ -90,7 +90,8 @@ const queueAutoSendInitials: Task = async (payload: Payload, helpers) => {
           null as queue_name,
           1 as max_attempts,
           now() + ((n / $2::float) * interval '1 second') as run_at,
-          1 as priority
+          -- prioritize in order as: autoassignment, autosending, handle delivery reports
+          4 as priority
         from contacts_to_queue
         -- this line doesn't modify the result at all, it just forces the execution of the intermediate CTE
         where id in ( select id from contacts_assigned )
--- a/src/server/tasks/queue-autosend-initials.ts
+++ b/src/server/tasks/queue-autosend-initials.ts
@@ -78,7 +78,7 @@ const queueAutoSendInitials: Task = async (payload: Payload, helpers) => {
         returning id
       ),
       new_jobs_queued as (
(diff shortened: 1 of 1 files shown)


src/server/tasks/queue-autosend-initials.ts
```diff
@@ -89,7 +89,8 @@ const queueAutoSendInitials: Task = async (payload: Payload, helpers) => {
           id::text as key,
           null as queue_name,
           1 as max_attempts,
-          now() + ((n / $2::float) * interval '1 second') as run_at
+          now() + ((n / $2::float) * interval '1 second') as run_at,
+          1 as priority
```
bchrobot, line 93: Suggestion: add a comment about jobs with a numerically smaller priority being run first

Suggestion: use a value of 10 to leave space for deprioritizing other tasks in the future without needing to move this one as well

bchrobot (approved):

src/server/tasks/queue-autosend-initials.ts
ajohn25: [Handling delivery reports](https://github.com/politics-rewired/Spoke/blob/main/migrations/20200804190149_deprioritize-handle-delivery-reports.js) is currently at 5 so can only go 1-4 for autosending to keep this behavior we want. I'll move to 4 if that sounds good? And I'll also comment on the 2 jobs the value was decided on relative to, to avoid the detective work 😅

bchrobot: Ah, gotcha - I think priority of 4 is good for now and then create a very low priority follow-up issue to create a strategy and documentation for task priority to use going forward (probably specifying priority tiers with `priority` in multiples of 100 along with some guidelines for how to think about priorities within Spoke Rewired).

PR #1442, opened 2022-09-28
feat(van): use real time sync

### LINKED ISSUE #1363: Rewrite VAN integration to use real-time sync

VAN's de-duplication logic is:

> It will attempt a match to the existing, but that's not a guaranteed match. ...the rule is that a match is attempted with results posted within the hour.

With our current ad-hoc sync approach, this results in duplicated canvass results each time a campaign is synced.

VAN also specifically requests real-time sync to distribute the load more evenly throughout the day.

Creating a question response record in Spoke should create a sync record to VAN with an associated task job.

This should ideally store a unique identifier from VAN for the canvass response, although [the docs](https://docs.ngpvan.com/reference/peoplevanidcanvassresponses) don't have a complete example response to confirm that's possible.

Editing or deleting a question response should attempt to update that response in VAN. The latest word from VAN suggestions that this is not yet possible:

> We do have functionality in testing that will allow applications with asynchronous conversations (like SMS threads) to update contact history records directly, but it's not ready for release yet.

N/A

**Additional context**

Sync-related performance issues were partially addressed in #1322.

hiemanshu:
## Description

Move to real time-ish VAN Sync. Still left to do for this PR:

- [x] Figure out optOut pathways 
- [x] Fill up the sync methods
- [x] Put functionality behind a config flag (until part 2 with UI changes are done)
- [ ] Figure out runAt time delay

## Motivation and Context

Fixes #1363 

## How Has This Been Tested?

Locally

NA. Non User Facing

## Checklist:


src/server/api/schema.js
```diff
@@ -193,10 +170,18 @@ async function deleteQuestionResponses(
...
-    .del();
+    .del()
+    .returning(["id"]);
+
+  for (const deletedQuestionResponse of deletedQuestionResponses) {
+    await queueExternalSyncForAction(
+      ActionType.QuestionReponse,
+      deletedQuestionResponse.id
+    );
+  }
```
bchrobot, line 184: Suggestion: use `Promise.all()`

```suggestion
  await Promise.all(deletedQuestionResponses.map(
    (deletedQuestionResponse) => queueExternalSyncForAction(
      ActionType.QuestionReponse,
      deletedQuestionResponse.id
    )
  ));
```

src/server/api/schema.js
```diff
@@ -1580,7 +1569,17 @@ const rootMutations = {
...
+        r.knex,
+        campaignContactId,
+        optOut,
+        loaders,
+        user
+      );
+
+      await queueExternalSyncForAction(ActionType.OptOut, result.optOutId);
+
+      return result.campaignContact;
```
bchrobot, line 1582: Question: the result type is changing here -- is that intentional?

src/server/models/cacheable_queries/opt-out.js
```diff
@@ -104,12 +105,16 @@ export const optOutCache = {
...
+      const [opttedOutId] = await trx("opt_out")
+        .insert({
+          assignment_id: assignmentId,
+          organization_id: organizationId,
+          reason_code: reason,
+          cell
+        })
+        .returning(["id"]);
+
+      optOutId = opttedOutId.id;
```
bchrobot, line 117: Suggestion: get rid of try/catch here and just use const for `optOutId`.

Almost four years later, I have no idea why I added squashing errors here: 995f4fe170d51b8368b12f46e37b2a3aefd79230

src/server/lib/external-systems.ts
```diff
@@ -1,40 +1,63 @@
...
-) => {
-  const [apiKey, existingMode] = van.api_key.split("|");
-  const mode = existingMode || DEFAULT_MODE;
-  request.auth(van.username, `${apiKey}|${mode}`, { type: "basic" });
-  request.url = `${config.VAN_BASE_URL}${request.url}`;
-  return request;
+  await r.knex.raw(
+    `
+  SELECT graphile_worker.add_job('queue-action-external-sync', ?)
+  `,
```
bchrobot, line 58: Nitpick:

```suggestion
    `select graphile_worker.add_job('queue-action-external-sync', ?)`,
```

src/server/lib/external-systems.ts
```diff
@@ -1,40 +1,63 @@
...
-
-export interface VanSecretAuthPayload {
-  username: string;
-  api_key: { __secret: string };
-}
+export const queueExternalSyncForAction = async (
+  actionType: ActionType,
+  actionId: number
+) => {
+  let campaign;
```
bchrobot, line 10: Required: type this

src/server/tasks/queue-action-external-sync.ts
```diff
@@ -0,0 +1,55 @@
...
+    })
+    .returning("id");
+
+  const externalSystem = await r
+    .knex("external_system")
+    .join("campaign", "campaign.external_system_id", "external_system.id")
+    .join("campaign_contact", "campaign_contact.campaign_id", "campaign.id")
+    .where({ "campaign_contact.id": campaignContactId })
+    .select(["external_system.id", "external_system.type"])
+    .first();
```
bchrobot, line 31: Required: only send the required data on the wire

```suggestion
    .first(["id", "type"]);
```

src/server/api/types.ts
```diff
@@ -291,3 +292,23 @@ export enum AutosendingControlsMode {
   Basic = "BASIC",
   Detailed = "DETAILED"
 }
+
+export interface IExternalSystem {
```
bchrobot, line 296: Required: drop the `I` to match existing Spoke conventions

```suggestion
export interface ExternalSystem {
```

src/server/external-systems/van.ts
```diff
@@ -0,0 +1,153 @@
...
+    const actionData = await r
+      .knex("action_external_system_sync")
+      .join(
+        "all_question_response",
+        "all_question_response.id",
+        "action_external_system_sync.action_id"
+      )
+      .join("all_external_sync_question_response_configuration", function () {
+        this.on(function () {
+          this.on(
```
bchrobot, line 95: Question is this extra layer of nesting required?

src/server/tasks/sync-contact-question-response.ts
```diff
@@ -0,0 +1,19 @@
...
+const syncContactQuestionResponse: Task = async (
+  payload: Record<string, any>,
+  helpers: JobHelpers
+) => {
+  switch (payload.externalSystemType) {
+    case ExternalSystemType.Van:
+      VAN.syncQuestionResponse(payload);
+      break;
+    default:
+      helpers.logger.error("Unsupported External System");
```
bchrobot, line 15: Question: do we want to squash the error like this? Same question for `syncContactOptOut()`. Update `sync_status` and `sync_error` here?

bchrobot (changes requested):
Looking good so far!

src/server/tasks/queue-action-external-sync.ts
hiemanshu: Already do that in the select. I'll remove the redundant select, and move the info needed into `first()`

src/server/external-systems/van.ts
hiemanshu: I think it's not required, knex suggests both ways of doing things here https://knexjs.org/guide/query-builder.html#join and I can't remember exactly why I went with the nested one here, I'll move to a non nested one and hope that nothing breaks.

src/server/tasks/sync-contact-question-response.ts
hiemanshu: Sounds good, will update the sync status and error!

hiemanshu:
@bchrobot @ajohn25 This is ready for review. Lots of changes in here. 

One flag is that some parts of the code as actually duplicates (some types mainly), and this is to prevent circular dependency issues. The code for VAN sync inside the tasks will eventually be removed, changes to it should ideally not happen as often, so we don't have to worry about updating multiple places either.

src/server/external-systems/van.ts
```diff
@@ -137,16 +335,165 @@ const VAN: IExternalSystem = {
...
+  ): Promise<void> {
+    const { campaignContactId, externalSystemId } = payload;
+
+    const campaignContact = await r
+      .knex("campaign_contact")
+      .where({ id: campaignContactId })
+      .first();
+
+    // Get all actions that need to be synced
+    // Ignores any that is not mapped
```
bchrobot, line 350: Question: should it be syncing all queued actions here? From the method name, `syncQuestionResponse`, I would expect it to sync just one.

My understanding is that creating an action will kick off a sync just that action, with a 10-15 minute debounce window to handle any short term corrections to that specific action.

src/server/external-systems/van.ts
```diff
@@ -137,16 +335,165 @@ const VAN: IExternalSystem = {
...
+    const contactCustomFields = JSON.parse(campaignContact.custom_fields);
+
+    const canvassResponse = {
+      canvassContext: {
+        phoneId: contactCustomFields.phone_id,
+        phone: formatPhone(campaignContact.cell),
+        contactTypeId: config.VAN_CONTACT_TYPE_ID,
+        dateCanvassed: DateTime.fromJSDate(
+          new Date(syncAction.created_at)
+        ).toFormat("MM-dd-yyyy")
```
bchrobot, line 467: Suggestion: skip JS Date intermediate:

```suggestion
        dateCanvassed: DateTime.fromSQL(syncAction.created_at).toFormat("MM-dd-yyyy")
```

src/server/lib/external-systems.ts
```diff
@@ -49,12 +57,12 @@ export const queueExternalSyncForAction = async (
...
     actionType,
     actionId,
-    campaignContactId: campaign.campaign_contact_id
+    campaignContactId: campaign?.campaign_contact_id
   };
 
   await r.knex.raw(
     `
-  SELECT graphile_worker.add_job('queue-action-external-sync', ?)
+  select graphile_worker.add_job('queue-action-external-sync', ?)
```
bchrobot, line 65: Nitpick: fix indentation. Either all on one line:

```ts
  await r.knex.raw(
    `select graphile_worker.add_job('queue-action-external-sync', ?)`,
    [payload]
  );
```

or balanced indentation:

```ts
  await r.knex.raw(
    `
      select graphile_worker.add_job('queue-action-external-sync', ?)
    `,
    [payload]
  );
```

src/config.js
```diff
@@ -396,6 +396,10 @@ const validators = {
     choices: ["hex"],
     default: "hex"
   }),
+  EXPERIMENTAL_VAN_SYNC: bool({
+    desc: "Use experimental real-time VAN sync",
+    default: false
+  }),
```
bchrobot, line 402: Question: should the standard sync be disabled when experimental sync is enabled?

src/server/external-systems/van.ts
```diff
@@ -36,7 +83,160 @@ export const withVan = (van: VanAuthPayload) => (
...
+        from message
+        where
+          campaign_contact_id = $1
+          and is_from_contact = false
+        order by id asc
+        limit 1
+      ),
+      canvass_responses as (
+        select
+          date_trunc('day', canvassed_at) as canvassed_at,
```
bchrobot, line 181: Required: handle date truncation as done in #1449

src/server/external-systems/van.ts
```diff
@@ -36,7 +83,160 @@ export const withVan = (van: VanAuthPayload) => (
...
+        join external_survey_question_response_option
+          on external_survey_question_response_option.id = qrro.external_response_option_id
+        join external_survey_question
+          on external_survey_question.id = external_survey_question_response_option.external_survey_question_id
+        where
+          external_survey_question.status = 'active'
+      ),
+      first_message as (
+        select
+          date_trunc('day', created_at) as canvassed_at,
```
bchrobot, line 168: Required: handle date truncation as done in #1449

bchrobot (changes requested):

src/server/external-systems/van.ts
hiemanshu: I envision 3 scenarios here:

1. Normal conversation, a question response is added, and no changes in the next 10 mins to this conversation. Single question response is synced
2. Super active conversation, multiple question responses are added, all actions are synced together 10 mins after last update. This prevents multiple jobs being setup, since VAN support batch inserts for multiple responses, this works well. 
3. Change by a sweeper or admin through message review, question responses were marked wrong, new actions are synced together 10 mins after the last update. Batch sync because it's supported by VAN. 

ActionKit or other external systems might not support batch sync, and since VAN supports it, I wanted to take advantage of it, and not sync one action per job which on an active campaign can potentially cause lots of performance issues to work through.

src/config.js
hiemanshu: Yeah. We can disable the "sync now" button without disabling the configure mapping functionality.

src/server/external-systems/van.ts
```diff
@@ -165,20 +171,20 @@ const getCanvassResponsesRaw = (
...
-          campaign_contact_id = $1
+          campaign_contact_id = ?
           and is_from_contact = false
         order by id asc
         limit 1
       ),
       canvass_responses as (
         select
-          date_trunc('day', canvassed_at) as canvassed_at,
+          date_trunc('day', canvassed_at at time zone (select timezone from cc_timezone)) at time zone (select timezone from cc_timezone) as canvassed_at,
```
bchrobot, line 187: Question: do we even need this with the new sync? Truncating by day was originally to allow grouping canvasses responses for batched upload. But with experimental sync, each is synced in realtime as a standalone canvass result and can use its own `created_at`

ajohn25 (approved):
Super readable for a big PR 🎉 Excited to finally know a little more about how the sync is / will be working under the hood after going over this 😁

src/api/external-system.ts
```diff
@@ -6,7 +6,8 @@ import type { ExternalResultCodeTarget } from "./external-sync-config";
 import type { RelayPaginatedResponse } from "./pagination";
 
 export enum ExternalSystemType {
-  VAN = "VAN"
+  VAN = "VAN",
+  DUMMY = "DUMMY"
```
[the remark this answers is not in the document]
hiemanshu: I added the dummy type to write a dummy implementation for this. I haven't implemented yet, and wanted to add a card. Thanks for checking, and I'll make a card! Adding a dummy implementation to allow adding other actions to be synced that other external systems (ActionKit, ActionNetwork), etc could support.

src/server/external-systems/van.ts
```diff
@@ -169,52 +158,26 @@ const getCanvassResponsesRaw = (
...
-          '[]'::json as result_codes,
-          '[]'::json as activist_codes,
-          '[]'::json as response_options
-        from message
-        where
-          campaign_contact_id = ?
-          and is_from_contact = false
-        order by id asc
-        limit 1
-      ),
```
bchrobot: Question: how are we accounting for a plain old "canvassed" action with this removed? Apologies if that's part of earlier diffs I've already reviewed!

e.g. the case where no question response was marked, no opt out recorded, etc.? Just "we sent them a message, mark it as  a canvass attempt"

bchrobot: Deferred to https://github.com/politics-rewired/Spoke/issues/1482

PR #1444, opened 2022-09-29
fix: reset dialog values when submitting

hiemanshu:
## Description

In opt outs import/export, the dialog values don't reset. If page is open, opt ins are imported, and you go back to import more, it keeps the previous file / list of pasted numbers.

Also keep export opt outs section open by default 

## Motivation and Context

Feedback from Organizing team

## How Has This Been Tested?

Tested Locally.

## Checklist:


ajohn25 (approved):

bchrobot (approved):
👍

PR #1445, opened 2022-09-29
fix(trollbot): fix token search

ajohn25:
## Description
This fixes filtering trollbot alarms by a specific token.
## Motivation and Context
Trollbot is significantly easier to use when token filtering is possible.

## How Has This Been Tested?
This has been tested locally.

## Checklist:


bchrobot (approved):
Question: was TypeScript previously mad about this? Is it now?

ajohn25:
> Question: was TypeScript previously mad about this? Is it now?

It was and is not after this change! Screenshot of the old error here
image

hiemanshu (approved):

Issue #1441, opened 2022-09-27
Allocate auto-send rate per-organization


bchrobot:
**Is your feature request related to a problem? Please describe.**

It is not possible to either a) limit auto-sending rate generally (to throttle for 10DLC, for example) or b) allocate throughput to specific organizations.

**Describe the solution you'd like**

Add a superadmin page to manage auto-sending rate allocation. A toggle at the top of the page could enable/disable per-organization rate limiting. When enabled, limits could be specified per-organization  (up to the instance-wide maximum of `AUTOSEND_MESSAGES_PER_SECOND`).

**Describe alternatives you've considered**

Just allowing superadmin to restrict instance-wide autosending rate below `AUTOSEND_MESSAGES_PER_SECOND`.

**Additional context**

N/A

bchrobot:
Feedback requested on this @ben-pr-p @hiemanshu @ajohn25

PR #1446, opened 2022-09-30
feat: add database backed per organization autosending

ben-pr-p:
## Description

This PR does two things:
- Changes autosending speed to be controlled by a database column that lives on the organization table
- Allows multiple campaigns to concurrently autosend across multiple organizations

## Motivation and Context

This accomplishes the following:
- Makes 10DLC compliance easier, since you can send at 30/sec on multiple campaigns
- Makes it easier for texters to simultaneously handle replies across multiple campaigns instead of sequentially dealing with big reply backlogs

See: https://github.com/politics-rewired/Spoke/issues/1441

## How Has This Been Tested?

Test suite

## Checklist:

### DIFF
--- a/__test__/testbed-preparation/core.ts
+++ b/__test__/testbed-preparation/core.ts
@@ -28,7 +28,11 @@ import {
 import type { HashedPassword } from "../../src/server/local-auth-helpers";
 
 export type CreateOrganizationOptions = Partial<
-  Pick<Organization, "name"> & { uuid: string; features: Record<string, any> }
+  Pick<Organization, "name"> & {
+    uuid: string;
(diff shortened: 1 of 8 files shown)


bchrobot (changes requested):
Required: while you're here update the job key in the `new_jobs_queued` CTE in `src/server/tasks/queue-autosend-initials.ts` to use some combination of task identifier, organization ID, and campaign contact ID to avoid collisions with other task types using keys.

Suggested: replace `insert into graphile_worker.jobs` with `select * from graphile_worker.add_job()`

bchrobot (approved):

PR #1449, opened 2022-10-05
fix: truncate day based on contact's timezone

bchrobot:
## Description

This truncates the day in the contact's timezone rather than UTC.

## Motivation and Context

Truncating to the day in UTC means it will always sync as the day before (e.g. 8:00pm the day before in ET).

It may be worth releasing this as an intermediate to #1442. Or it may not!

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/server/tasks/ngp-van/sync-campaign-contact-to-van.ts
+++ b/src/server/tasks/ngp-van/sync-campaign-contact-to-van.ts
@@ -85,7 +85,13 @@ export const fetchCanvassResponses = async (
   client
     .query<CanvassResultRow>(
       `
-        with configured_response_values as (
+        with cc_timezone as (
+          select coalesce(cc.timezone, c.timezone) as timezone
+          from campaign_contact cc
+          join campaign c on c.id = cc.campaign_id
+          where cc.id = $1
+        ),
+        configured_response_values as (
           select
             qrc.id,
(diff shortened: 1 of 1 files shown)


src/server/tasks/ngp-van/sync-campaign-contact-to-van.ts
```diff
@@ -153,7 +159,7 @@ export const fetchCanvassResponses = async (
         ),
         first_message as (
           select
-            date_trunc('day', created_at) as canvassed_at,
+            date_trunc('day', created_at at time zone (select timezone from cc_timezone)) at time zone (select timezone from cc_timezone),
```
hiemanshu, line 162: Question: do these need an `as canvassed_at` ?

bchrobot: I missed this, sorry! I think they do because of this below:

```ts
  const {
    canvassed_at: dateCanvassed,
    response_options,
    activist_codes,
    result_codes
  } = canvassResultRow;
```

bchrobot: I can open a PR to fix

PR #1468, opened 2022-10-10
fix: media attachment false positives

### LINKED ISSUE #1102: Media attachment error message false positives
It looks like browser-based checks will not work for this and we'll need to add a resolver for performing this validation server-side (see comment below). Open to suggestions, however.

Media attachment in question: `https://s3.amazonaws.com/ak-newsom/images/GGN_P2P_Image_220131_v1.jpeg`

We should also check the image's size while we're here.

hiemanshu:
## Description

CORS could block images from being downloaded to test for MIME types, so download them in the backend, and use libmagic to test for the file mime type

## Motivation and Context

Fixes #1102 

## How Has This Been Tested?

Tested locally with multiple files

## Checklist:


src/server/api/schema.js
```diff
@@ -4029,6 +4037,20 @@ const rootResolvers = {
           count: result.count
         };
       });
+    },
+    isValidAttachment: async (_root, { fileUrl }, _context) => {
+      const filePath = path.join(
```
bchrobot, line 4042: Required: this file should be deleted in a `finally` block of a try/catch/finally statement.

Suggestion: perhaps a utility like below would be good for this?

```ts
type TempDownloadHandler = <T extends unknown>(filePath: string) => Promise<T> | T;

const withTempDownload = async (url: string, handler: TempDownloadHandler) => {
  const tmpPath = ...;

  try {
    // download to tmpPath
    const result = await handler(tmpPath);
    return result;
  } finally {
    os.rm(tmpPath);
  }
};
```

src/server/api/lib/file-type.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const getFileType = async (filePath: string) => {
+  FileMagic.magicFile = require.resolve("@npcz/magic/dist/magic.mgc");
+
+  if (process.platform === "darwin" || process.platform === "linux") {
+    FileMagic.defaulFlags = MagicFlags.MAGIC_PRESERVE_ATIME;
+  }
+
+  const magic = await FileMagic.getInstance();
+  const fileType = magic.detect(
```
bchrobot, line 13: Comment: a little bit of a bummer `magic` doesn't accept streams or buffers

src/lib/utils.ts
```diff
@@ -0,0 +1,101 @@
+import fs from "fs";
+import escapeRegExp from "lodash/escapeRegExp";
+import isEqual from "lodash/isEqual";
+import isObject from "lodash/isObject";
+import transform from "lodash/transform";
+import request from "request";
```
bchrobot, line 6: Required: `superagent` instead of `request` for consistency with all post-fork code.

bchrobot (changes requested):

bchrobot (approved):

Issue #1146, opened 2022-04-12
Add conventions document


bchrobot:
**Is your feature request related to a problem? Please describe.**

It is difficult to work on a large codebase with missing or competing conventions.

We have prettier to enforce a single code format and eslint to enforce some style choices/conventions but they do not capture everything.

**Describe the solution you'd like**

Add a `conventions.md` document with codebase conventions and style guidelines to capture things that prettier and eslint do not.

**Describe alternatives you've considered**

Custom eslint rules.

**Additional context**

Running list of things to include:

* Component props interface names should be `[ComponentName]Props`, not just `Props`.

bchrobot:
@politics-rewired/spoke-rewired thoughts on what else to include in the first draft of this?

bchrobot:
How we handle GraphQL `errors` under different circumstances.

- fetching data for display: inline list of errors, using MUI [`<Alert>` component](https://v4.mui.com/components/alert/#description)
- performing an operation from a button -- modal with error? MUI [`<Snackbar>` component](https://v4.mui.com/components/snackbars/#customized-snackbars)?
- performing an operating from within a modal -- inline errors at top of modal?

ajohn25:
> How we handle GraphQL `errors` under different circumstances.
> 
> * fetching data for display: inline list of errors, using MUI [`<Alert>` component](https://v4.mui.com/components/alert/#description)
> * performing an operation from a button -- modal with error? MUI [`<Snackbar>` component](https://v4.mui.com/components/snackbars/#customized-snackbars)?
> * performing an operating from within a modal -- inline errors at top of modal?

Just mentioning #1270 here to link the discussions together

ajohn25:
- Up to a single style for inline CSS - otherwise use `makeStyles` hooks. See #1610
- Affirmative variables for component behavior (show, enable, etc.). See #1606

PR #1473, opened 2022-10-17
feat: support per-campaign autosend limit

### LINKED ISSUE #1318: Autosending contact limit configurable by campaign
In order to send a test to a limited number of recipients, admins must load two campaigns: one with the test universe and another with the full universe. If further testing is needed, they are required to START the full universe for autosending, monitor it, and stop it after a desired amount of contacts are sent. This prevents autosending initials from being a hands-off endeavor — No good!

From the autosending page and for each campaign, a field should exist next to _Autosending Status_ where:
* If left blank, the entire campaign should autosend unless interrupted
* If given a limit (e.g. 20,000 contacts), autosending should pause sending after 20,000 contacts and _Autosending Status_ tag should read `Paused`; limit field should retain `20,000`
* If a limit is given and reached (e.g. 20,000 contacts), the admin should be able to remove the limit and hit START in order to send to the full universe
* If limit is modified after autosending has started but before that number of contacts has been sent an initial message, the new limit should override the previous and no more than that TOTAL NUMBER should receive an initial message until the limit is removed
* If a limit is modified after autosending has started but after that number of contacts has been sent an initial message, the field should disallow the admin to enter a number that is lower than the original limit

'Autosending Status' could read something other than PAUSED when limit is reached, e.g. LIMIT

bchrobot:
## Description

This allows autosend admins to specify a contact count limit for autosending. When set, the first _n_ messages will go out and then the campaign will pause.

## Motivation and Context

Closes #1318.

## How Has This Been Tested?

See tests added in PR.

<table><tr><td>

Screenshot 2022-10-17 at 12 43 17 PM

_Save on type_

</td></tr><tr><td>

Screenshot 2022-10-17 at 6 04 58 PM

_Server error_

</td></tr></table>

## Documentation Changes

N/A

## Checklist:

---
  - https://app.asana.com/0/0/1202577617170592


src/server/api/schema.js
```diff
@@ -1979,6 +1979,50 @@ const rootMutations = {
...
+                from (
+                  select id
+                  from campaign_contact
+                  where campaign_id = ?::int
+                  order by id asc
+                  limit ?::int
+                ) campaign_contact_ids
+              )
+              else null
+            end)
```
ben-pr-p, line 2016: Is the reason we handle the case where limit is null so that this query can double for setting autosend_limit/max_contact_id to null?

nit: I think it'd be easier to read as 2 separate queries

src/server/tasks/queue-autosend-initials.ts
```diff
@@ -168,45 +171,57 @@ export const queueAutoSendOrganizationInitials: Task = async (
...
-        -- campaign requirements for autosending
-        and c.is_archived = false
-        and c.is_started = true
-        and c.autosend_status = 'sending'
-        and message_status = 'needsMessage'
-        and is_opted_out = false
-      group by 1
+        and organization_id = $1
+        and campaign_summary.campaign_id = campaign.id
+        and new_autosend_status is not null
```
ben-pr-p, line 223:
```suggestion
        and new_autosend_status is not null
        and id = ANY($2::integer[])
```
We only want to be able to mark campaigns as complete that we just queued, right? That's preserving previous behavior

ben-pr-p (approved):

src/server/tasks/queue-autosend-initials.ts
bchrobot: The limit to `$2` campaign IDs passes through from `campaign_summary_raw` but I'm happy to duplicate it in the `update` as a safety

src/server/api/schema.js
bchrobot: Yeah, that was the reason. I've split it into two queries.

src/server/tasks/queue-autosend-initials.ts
bchrobot: Done in https://github.com/politics-rewired/Spoke/pull/1473/commits/a799bb901459b399724e21e6302e8b235cd3f714

kohlee:
This has been tested in production and we'd like to prioritize its release!

migrations/20221016075536_add_autosend_limit.js
```diff
@@ -0,0 +1,295 @@
...
+        where is_template = false;
+      `);
+    });
+};
+
+exports.down = function down(knex) {
+  return knex.schema
+    .raw(
+      `
+      drop view campaign cascade;
```
hiemanshu, line 52: comment: I hate how much work this requires. abysmal.

src/containers/AdminAutosending/components/AutosendingLimitField.tsx
```diff
@@ -0,0 +1,83 @@
...
+import ErrorIcon from "@material-ui/icons/Error";
+import {
+  useGetCampaignAutosendingLimitQuery,
+  useUpdateCampaignAutosendingLimitMutation
+} from "@spoke/spoke-codegen";
+import isNil from "lodash/isNil";
+import React, { useCallback, useMemo, useState } from "react";
+import { useDebouncedCallback } from "use-debounce";
+
+export interface AutosendingLimitFieldProps {
```
hiemanshu, line 15: nitpick: no export.

src/server/api/schema.js
```diff
@@ -1979,6 +1979,62 @@ const rootMutations = {
...
+        .where({ id })
+        .first(["organization_id"]);
+
+      const organizationId = campaign.organization_id;
+      await accessRequired(user, organizationId, "ADMIN", true);
+
+      const {
+        rows: [updatedCampaign]
+      } =
+        limit === null
```
hiemanshu, line 2000: nitpick / suggestion: use knex methods to write smaller queries.

hiemanshu (approved):
LGTM! A few nitpicks / suggestions.

migrations/20221016075536_add_autosend_limit.js
bchrobot: Yep :(

I would back a proposal for Rewired to sponsor development of an `alter view drop column`. There's been discussion about [this since 2008](https://www.postgresql.org/message-id/603c8f070807291912x37412373q7cd7dc36dd55a8a7@mail.gmail.com).

src/containers/AdminAutosending/components/AutosendingLimitField.tsx
bchrobot: Curious why no export here?

The props for the component are a public interface and what I've seen in other packages is exporting the props types along with the component.

src/server/api/schema.js
bchrobot: Done in https://github.com/politics-rewired/Spoke/pull/1473/commits/2972d9938243a0426d499187dbc2821ef6c5a5f7

src/containers/AdminAutosending/components/AutosendingLimitField.tsx
hiemanshu: Most other components we don't export the props interface, only the component with a default export. So it's a nitpick for me for not keeping with the standard convention. If we want to export props, we should make that the convention.

src/server/api/schema.js
bchrobot: You are correct. I retested locally and it did not work.

Fixed and confirmed in https://github.com/politics-rewired/Spoke/pull/1473/commits/2972d9938243a0426d499187dbc2821ef6c5a5f7

src/containers/AdminAutosending/components/AutosendingLimitField.tsx
bchrobot: I think exporting the props should be the convention, yeah. Added here:

https://github.com/politics-rewired/Spoke/issues/1146#issuecomment-1284277601

Issue #1344, opened 2022-08-09
Separate Redis caching into 4 tiers which can be independently enabled


ben-pr-p:
Right now, we have an implementation of a basic cache + invalidate system with our Memoredis system.

This system is good, but we're not 100% confident that we are invalidating it in every circumstance that we should be.

However, we can still get a good chunk of the caching benefits by dividing `cacheOpts` into 4 buckets.

1. `core` - this bucket is all of the things that are included on every single HTTP request (authentication, user_organization status). These are never changed via Retool or direct database queries, and we can be pretty confident that our system is accurate here. These should be enabled for every single client.
2. `aggregates` - this bucket is for all of the things that we don't need perfect freshness for, but where freshness is nice (top level campaign stats, etc.). This bucket can be enabled without negative consequence for our biggest users, and we can not enable it for our smaller ones since their aggregates are performant.
3. `advanced` - this bucket contains more complex things like `CampaignTeams`, `CampaignsList`, `CampaignsListRelay`. These are things we are not that confident we don't change inside the database, and 
4. `conversation` - this bucket contains basically just this block (things that are requested very frequently, we do change via Retool, but we still have significant performance gains by caching these for every 15 - 30 seconds). 
```
  CampaignInteractionSteps: ["campaign-interaction-steps", THIRTY_SECONDS],
  CampaignCannedResponses: ["campaign-canned-responses", THIRTY_SECONDS],
  InteractionStepChildren: ["interaction-step-children", THIRTY_SECONDS],
  InteractionStepSingleton: ["interaction-step-singleton", THIRTY_SECONDS],
```

We should be able to control these via a new comma separated environment variable (`MEMOREDIS_ACTIVE_CACHE_BUCKETS=core,aggregates`). When the environment variable doesn't include the bucket, `memoredis.invalidate` does nothing, and the function that memoredis wraps is just called without checking Redis.

bchrobot:
@ben-pr-p how are you imagining `MEMOREDIS_ACTIVE_CACHE_BUCKETS` interacting with `memoredis`? It seems like something best passed to the [`memoredis` constructor](https://github.com/politics-rewired/memoredis#initialization).

hiemanshu:
## Hosting 

Hosting considerations: 12GB (RAM) Usable Storage, 1 Main + 1 Read Replica, Snapshotting disabled (we don’t store critical data, only as a caching layer)

[...]

PR #1483, opened 2022-10-21
feat: add redis caching layer, implement core

hiemanshu:
core includes caching campaigns lists, and organization

## Description

This implements memoredis based caching / invalidation for campaigns list, and organization data. 

## Motivation and Context

Part 1 of fixing #1344, needs part 2 to add support for buckets, early PR to help with mlabs 

## How Has This Been Tested?

Tested locally. 

## Checklist:


hiemanshu:
Still need to fix: 

- `GraphQLDate` doesn't work with `JSON.parse`d `createdAt`
- Wait for v2 memoredis to be up in npm repo, and upgrade to it

src/server/contexts/index.ts
```diff
@@ -2,32 +2,20 @@
...
-
-const createContext = (host: string): SpokeContext => ({
+const createContext = (_host: string): SpokeContext => ({
   db: {
     schema: "public",
     primary: r.knex,
     reader: r.reader
   },
-  memoizer: createHostMemoizer(host)
+  memoizer: createMemoizer({ emptyMode: true }).then((memoizer: Memoizer) => {
```
bchrobot, line 16: Question: why force `emptyMode` even when `CACHING_URL` is present?

src/server/memoredis.ts
```diff
@@ -20,59 +10,121 @@ const ONE_HOUR = ONE_MINUTE * 60;
...
+  key: string;
+  ttl: number;
 };
 
-const cacheOpts = Object.fromEntries(
-  Object.entries(cacheOptsPlain).map(([name, [key, ttl]]) => [
-    name,
-    { key, ttl }
-  ])
-);
```
bchrobot, line 76: ❤️

src/server/api/campaign.js
```diff
@@ -139,7 +139,7 @@ export const resolvers = {
   },
   CampaignStats: {
     sentMessagesCount: async (campaign) => {
-      const getSentMessagesCount = memoizer.memoize(async ({ campaignId }) => {
+      const getSentMessagesCount = async ({ campaignId }) => {
```
bchrobot, line 142: Comment for future work on this: can we just get counts grouped by status in one cached query and then do addition ourselves for each type of `count...` resolver?

bchrobot (changes requested):
One requested change but otherwise looks good!

I kept a running list places where caching was removed as I was going through everything anyway:

`src/server/api/assignment.js`:
- `memoizedMyCurrentAssignmentTargets`
- `notifyIfAllAssigned`
- `doAssignment`
- `getCampaign`
- `getCannedResponses`

`src/server/api/campaign.js`:
- ~~`doGetCampaigns`~~ (cached inline)
- `getCampaignOrganization`
- `getSentMessagesCount`
- `getReceivedMessagesCount`
- `getOptOutsCount`
- `getNeedsMessageOptOutsCount`
- `getCountMessagedContacts`
- `getCountNeedsMessageContacts`
- `getPercentUnhandledReplies`
- `getCampaignTeams`
- `getInteractionSteps`
- `getCannedResponses` (can this be shared with the `getCannedResponses` in `assignment.js`?)
- `getHasUnsentInitialMessages`
- `getHasUnhandledMessages`

`src/server/api/errors.js`:
- `getUserRole`

`src/server/api/lib/campaign.ts`:
- various invalidatons in `editCampaign`

`src/server/api/lib/interaction-steps.ts`:
- various invalidatons in `persistInteractionStepNode`

`src/server/api/organization.js`:
- `getTags`
- `getEscalationTags`

`src/server/api/question.js`:
- `getAnswerOptions`
- `getInteractionStep`

`src/server/api/schema.js`:
- various invalidatons

`src/server/api/user.js`
- `getOrganizationsForUserWithRole`
- `getRoleForUserInOrganization`

`src/server/models/cacheable_queries/user.ts`
- `getUserByAuth0Id`
- `getUserById`

src/server/contexts/index.ts
hiemanshu: Since `createMemoizer` is async, without `emptyMode` it doesn't return quick enough to have a useable interface, this needs to be fixed before final release.

hiemanshu:
This PR is almost ready for review. @bchrobot @ben-pr-p 

Left to do:
1. Add a custom `GraphQLDate` time that uses DateTime
2. Figure out context cache. 

Neither of these are required to be done in this PR. So we can merge change as is, and come back to these later as well.

src/server/contexts/index.ts
bchrobot: The `ApolloServer` constructor takes an async function for `context`:

https://github.com/politics-rewired/Spoke/blob/9d0b9267e3f8abbe01cb6cf784f528427e4b9acc/src/server/routes/graphql.ts#L61

so you should be able to just add async/await for each method requiring it

bchrobot (approved):

src/server/api/campaign.js
```diff
@@ -110,10 +110,9 @@ const doGetCampaigns = async ({ organizationId, cursor, campaignsFilter }) => {
 
 export async function getCampaigns(organizationId, cursor, campaignsFilter) {
   const memoizer = await MemoizeHelper.getMemoizer();
-  const memoizedCampaigns = memoizer.memoize(
-    doGetCampaigns,
-    cacheOpts.CampaignsList
-  );
+  const memoizedCampaigns = MemoizeHelper.hasBucketConfigured(Buckets.Advanced)
+    ? memoizer.memoize(doGetCampaigns, cacheOpts.CampaignsList)
```
bchrobot, line 114: Suggestion: I think the bucket check should be handled within `memoizer.memoize` in the same way that the Redis client (via presence of Redis URI envvar) is checked within that. The bucket could become part of the `cacheOpts` entry.

src/server/api/schema.js
```diff
@@ -849,11 +849,6 @@ const rootMutations = {
         })
         .returning("*");
 
-      const memoizer = await MemoizeHelper.getMemoizer();
-      await memoizer.invalidate(cacheOpts.CampaignsList.key, {
-        organizationId: organization.id
-      });
```
bchrobot: Question: I think we want to keep this invalidation? A campaign has just been inserted so the campaign list result will change.

src/server/api/schema.js
```diff
@@ -1065,11 +1060,6 @@ const rootMutations = {
         );
       }
 
-      const memoizer = await MemoizeHelper.getMemoizer();
-      await memoizer.invalidate(cacheOpts.CampaignsList.key, {
-        organizationId: origCampaign.organization_id
-      });
```
bchrobot: Question: I think we want to keep this invalidation? A campaign has just been edited so the campaign list result may have changed.

src/server/api/schema.js
hiemanshu: We're doing it inside `editCampaign` instead now! https://github.com/politics-rewired/Spoke/blob/6088c8a41b324ac48c0d28beb251309065a030e4/src/server/api/lib/campaign.ts#L736

This should fix the invalidate happening when something happens, and then someone else possibly calling the campaigns list, and us getting back stale campaigns, instead the invalidate now happens at the end of `editCampaign` since we do a `return editCampaign` in both scenarions (this and below).

bchrobot (approved):

PR #1503, opened 2022-10-30
refactor(autosending): calculate limit at execution time

### LINKED ISSUE #1491: Improve performance of campaign contact deletion
#1473 added an unindexed foreign key from campaign to campaign contact.

This dramatically slows down deletes from the `campaign_contact` table. Deleting ~70,000 rows takes <400ms for the actual table operation but ~27,000ms to check the foreign keys.

Proposed solutions:
1. Index `campaign.autosend_limit_max_contact_id`
2. Drop `campaign.autosend_limit_max_contact_id` and calculate at runtime based on `autosend_limit`. This could be cached NodeJS-side and invalidated any time campaign contacts are deleted.

bchrobot:
## Description

This refactors `queue-autosend-organization-initials` to perform limits at runtime rather than pre-computing.

## Motivation and Context

Closes #1491.

Also fixes incompatibility between current autosend limit implementation and prioritization of unassigned conversations. 

## How Has This Been Tested?

This has been tested locally. Also with test suite.

Down migration has been verified by adding `yarn knex migrate:down` to the end of `update-dump.sh` and ensuring the diff for `schema-dump.sql` is empty.

N/A

## Documentation Changes

N/A

## Checklist:

---
  - https://app.asana.com/0/0/1203228252450339


migrations/20221030115356_dynamic-autosend-limit.js
```diff
@@ -0,0 +1,365 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
```
ajohn25, line 4: Question: Are these summary comments autogenerated now?

src/server/tasks/queue-autosend-initials.ts
```diff
@@ -171,58 +180,34 @@ export const queueAutoSendOrganizationInitials: Task = async (
...
         select
           c.id as campaign_id,
-          c.autosend_limit_max_contact_id,
-          count(cc.id) as total_count_to_send,
-          min(cc.id) as min_cc_id
+          count(cc.id) filter (where message_status <> 'needsMessage') as sent_count,
+          count(cc.id) filter (where message_status = 'needsMessage') as to_send_count
         from campaign c
-        left join sendable_contacts cc on cc.campaign_id = c.id
+        left join campaign_contact cc on cc.campaign_id = c.id
```
ajohn25, line 191: Nitpick: Can this be an inner join now?

ajohn25 (approved):

hiemanshu (approved):
LGTM! Couple of suggestions from Aashish, and need main merged in!

migrations/20221030115356_dynamic-autosend-limit.js
bchrobot: They are, yeah

src/server/tasks/queue-autosend-initials.ts
bchrobot: We want it to be a left join on the off-chance there is a campaign in `sending` status that doesn't have contacts. It's unlikely but could happen if the campaign is last to be queued on a long-ish list and, after it's start/queuing someone runs "delete unmessaged".

Issue #1398, opened 2022-08-31
Refactor schema.js


hiemanshu:
`schema.js` is way too large and unwieldy standing at around just over 4k lines. Refactor schema.js over multiple parts

### Part 1: 
[ ] Break `schema.js` into two files, `rootMutations` and `rootResolvers`. Copy over used functions into their respective files
[ ] Save `rootMutations` and `rootResolvers` as `.ts` files. And make lint happy for now. 

### Part 2:
[ ] Refactor `rootMutations` to move out large mutating functions into libraries in the `./lib/` subfolder (`src/server/api/lib`) adding type definitions as you go along. 
[ ] Refactor `rootResolvers` to move functions into libraries in the `./lib/` sub folder adding type definitions as you go along.

At the end of part 2 `schema.js` will mostly be a file stitching all the queries or mutations together, and the root mutations and root queries will be just stitching different parts of the library (`./lib/` subfolder) together. Most of the code will live in individually namespaced (and easy to spec) libraries that we can add tests to more easily.

bchrobot:
@hiemanshu what do you have in mind for making TS compiler happy about js -> ts in Step 1? Switching to TS and adding correct typedefs seem to me like they go better together and fit better in Part 2.

hiemanshu:
> @hiemanshu what do you have in mind for making TS compiler happy about js -> ts in Step 1? Switching to TS and adding correct typedefs seem to me like they go better together and fit better in Part 2.

I think we can ignore making TS Compiler happy in the step 1, just want to make lint happy initially. Renaming it to `.ts` will force any new changes being added to it to be made with type defs, etc.

PR #1510, opened 2022-10-31
refactor: schema.js into resolvers and mutations file

hiemanshu:
## Description

Refactor schema.js into two files, resolvers and mutations. New changes being added into their own files and most code should live in `src/api/lib` 

## Motivation and Context

This is part 1 of #1398 

## Checklist:


bchrobot (approved):
Looks good to me!

I imagine both new `.ts` files are riddled with "no implicit any" warnings?

src/lib/utils.ts
```diff
@@ -8,7 +8,7 @@ import os from "os";
 import path from "path";
 import request from "superagent";
 
-export type TempDownloadHandler = <T>(filePath: string) => Promise<T> | T;
+export type TempDownloadHandler = (filePath: string) => Promise<any> | any;
```
ajohn25, line 11: Is this change supposed to be part of this PR?

ajohn25 (approved):

hiemanshu:
> Looks good to me!
> 
> I imagine both new `.ts` files are riddled with "no implicit any" warnings?

Yup. Filled with them, but it's a start, we can refactor as we go on, without having to cause large disruptions. Lint doesn't complain, only editors, so something we can live with, and start working towards it!

src/lib/utils.ts
hiemanshu: This change is part of this PR. I started fixing small type issues I could find. Then thought I'd just get this PR'd first, and can start adding types as time goes on. The longer this split would sit in flux, the longer it would cause pain/disruptions.

PR #1518, opened 2022-11-04
feat(assemble-numbers): add counter to sendMessage

ben-pr-p:
## Description

This adds a counter to send message calls for more granular monitoring and detection.

It appears to be a bit difficult to distinguish between HTTP level errors and 200 + GraphQL errors right now, so we can punt that for now. This accomplishes the main goal of enabling better alerting, and then we can turn to logs to further diagnose in the event of an alert.

## How Has This Been Tested?

TODO

### DIFF
--- a/src/server/api/lib/assemble-numbers.ts
+++ b/src/server/api/lib/assemble-numbers.ts
@@ -1,11 +1,10 @@
-import type { PoolClient } from "pg";
-
 import { config } from "../../../config";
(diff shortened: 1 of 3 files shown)


hiemanshu (approved):
LGTM! For the TODO, we should create a new issue, so it's not lost just a comment.

ben-pr-p:
Created the issue!

bchrobot (approved):

PR #1532, opened 2022-11-17
fix(opt outs): add unique cell per organization constraint

ajohn25:
## Description

This PR:
1. Adds a unique cell per organization constraint to the `opt_out` table
2. Adds checking for an existing opt out before processing a new auto opt out in `message-sending.js`

## Motivation and Context

Having multiple opt out records for the same conversation can be confusing for read replica users who expect at most 1 record to be returned per campaign contact when joining to `opt_out`. Duplicate records are currently created for example if a contact spams "STOP" as a reply.

## How Has This Been Tested?
This has been tested locally.

## Checklist:

### DIFF
--- a/src/server/api/lib/message-sending.js
+++ b/src/server/api/lib/message-sending.js
@@ -392,8 +392,6 @@ export async function saveNewIncomingMessage(messageInstance) {
   let updateQuery = r.knex("campaign_contact").limit(1);
 
   if (OPT_OUT_TRIGGERS.includes(cleanedUpText)) {
-    updateQuery = updateQuery.update({ message_status: "closed" });
-
     const { id: organizationId } = await r
       .knex("organization")
       .first("organization.id")
@@ -401,14 +399,25 @@ export async function saveNewIncomingMessage(messageInstance) {
       .join("assignment", "campaign_id", "=", "campaign.id")
       .where({ "assignment.id": assignment_id });
 
-    const optOutId = await cacheableData.optOut.save(r.knex, {
-      cell: contact_number,
-      reason: "Automatic OptOut",
-      assignmentId: assignment_id,
-      organizationId
-    });
+    const optOutQuery = r
+      .knex("opt_out")
+      .first("id")
+      .where({ cell: contact_number });
 
-    await queueExternalSyncForAction(ActionType.OptOut, optOutId);
+    const existingOptOut = await optOutQuery;
+
+    if (existingOptOut === undefined) {
+      updateQuery = updateQuery.update({ message_status: "closed" });
+
+      const optOutId = await cacheableData.optOut.save(r.knex, {
+        cell: contact_number,
+        reason: "Automatic OptOut",
+        assignmentId: assignment_id,
+        organizationId
+      });
+
+      await queueExternalSyncForAction(ActionType.OptOut, optOutId);
+    }
   } else {
     updateQuery = updateQuery.update({ message_status: "needsResponse" });
   }
--- a/migrations/20221117021725_remove-duplicate-opt-outs.js
+++ b/migrations/20221117021725_remove-duplicate-opt-outs.js
@@ -1,0 +1,19 @@
+exports.up = function up(knex) {
+  return knex.raw(`
+    delete from opt_out oo
+    using opt_out existing
+    where oo.id > existing.id
+      and oo.cell = existing.cell
+      and oo.organization_id = existing.organization_id;
+
+    alter table opt_out
+      add constraint unique_cell_per_organization_id unique (cell, organization_id)
+  `);
+};
+
+exports.down = function down(knex) {
+  return knex.raw(`
+    alter table opt_out
+      drop constraint unique_cell_per_organization_id
+  `);
+};
--- a/schema-dump.sql
+++ b/schema-dump.sql
@@ -4210,6 +4210,14 @@ ALTER TABLE ONLY public.unhealthy_link_domain
 
 
 --
+-- Name: opt_out unique_cell_per_organization_id; Type: CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY public.opt_out
+    ADD CONSTRAINT unique_cell_per_organization_id UNIQUE (cell, organization_id);
+
+
+--
 -- Name: canned_response unique_per_campaign; Type: CONSTRAINT; Schema: public; Owner: postgres
 --
 


src/server/api/lib/message-sending.js
```diff
@@ -392,23 +392,32 @@ export async function saveNewIncomingMessage(messageInstance) {
...
       .join("assignment", "campaign_id", "=", "campaign.id")
       .where({ "assignment.id": assignment_id });
 
-    const optOutId = await cacheableData.optOut.save(r.knex, {
-      cell: contact_number,
-      reason: "Automatic OptOut",
-      assignmentId: assignment_id,
-      organizationId
-    });
+    const optOutQuery = r
```
hiemanshu, line 402: Required: We should move these checks into `optOut.save` method. We use this method in many places, and we could either add checks in every place, or just update the save method to perform the check instead.

hiemanshu (changes requested):

src/server/api/lib/message-sending.js
ajohn25: Looking more into this I think we have to do it in each place separately because there's other actions that also shouldn't be repeated if we already have an opt out record (ex. closing convo and queuing external sync here). I found 2 of those:

1. [`createOptout`](https://github.com/politics-rewired/Spoke/blob/main/src/server/api/root-mutations.ts#L184): My understanding is that this only gets called from a texter assignment or message review, so I think rather than check for an existing opt out here, we do `optOut.save` in a try/catch with a log. That would be similar to us logging errors when someone tries to send an initial to an opted out contact.

2. `optOut.saveMany`: To still take advantage of doing a bulk insert, I think we just modify the insert query  to add an `on conflict do nothing`. In this case we don't need to keep track of the error because we expect that users will upload numbers they've already opted out sometimes.


Does ^ sound right to you?

hiemanshu: Yeah. Sounds good to me! With knex 2.x you can do an `onConflict(...).update(timestamp)` so that the `updated_at` gets updated.

ajohn25: Would be so nice, exceptt we're using the rethink adapter here so can't even call `onConflict()` 😞

bchrobot: `opt_out.updated_at` already gets updated `BEFORE UPDATE` via a Postgres trigger.

Question: what is the value of updating `updated_at` during a conflict versus just ignoring the row?

src/server/api/root-mutations.ts
```diff
@@ -181,12 +181,17 @@ async function createOptOut(trx, campaignContactId, optOut, loaders, user) {
...
+  try {
+    optOutId = await cacheableData.optOut.save(trx, {
+      cell,
+      reason,
+      assignmentId,
+      organizationId
+    });
+  } catch (error) {
+    logger.error("Error creating opt out:", error);
+  }
```
bchrobot, line 194: Question: why are we squashing the error here? If it's a Postgres error the transaction will be tainted and the rest of the SQL statements using the transaction will fail.

If just for logging then we want to re-throw the error. We'll also want to convert the error to JSON with `errToObj()`

```suggestion
  } catch (error) {
    logger.error("Error creating opt out:", errToObj(error));
    throw error;
  }
```

src/server/models/cacheable_queries/opt-out.js
```diff
@@ -151,7 +151,10 @@ export const optOutCache = {
         reason_code: reason
       }));
 
-      await trx("opt_out").insert(optOuts);
+      // workaround for rethink knex adapter using knex 0.17
```
bchrobot, line 154: Question: can we force rethink to use the latest knex using [the `resolutions` field of `package.json`](https://classic.yarnpkg.com/lang/en/docs/package-json/#toc-resolutions)?

src/server/api/lib/message-sending.js
```diff
@@ -392,23 +392,32 @@ export async function saveNewIncomingMessage(messageInstance) {
...
-      reason: "Automatic OptOut",
-      assignmentId: assignment_id,
-      organizationId
-    });
+    const existingOptOutQuery = r
+      .knex("opt_out")
+      .first("id")
+      .where({ cell: contact_number });
+
+    const existingOptOut = await existingOptOutQuery;
```
bchrobot, line 407: Required: combine these to just await the query. We don't conditionally modify the query or clone it so we don't need to keep a reference to the `QueryBuilder` instance.

src/server/api/lib/message-sending.js
```diff
@@ -392,23 +392,32 @@ export async function saveNewIncomingMessage(messageInstance) {
...
+      const optOutId = await cacheableData.optOut.save(r.knex, {
+        cell: contact_number,
+        reason: "Automatic OptOut",
+        assignmentId: assignment_id,
+        organizationId
+      });
 
-    await queueExternalSyncForAction(ActionType.OptOut, optOutId);
+      await queueExternalSyncForAction(ActionType.OptOut, optOutId);
+    }
```
bchrobot, line 420: Question: it seems like we would want to mark the conversation as closed regardless of whether there's an existing opt out? Similarly, it seems like we would want to queue an external sync for _any_ opt out event, even if that contact is already opted out in Spoke's system?

If so, those two could be pulled up and the existingOptOut check could be replaced with an `on conflict do ...` in `cacheableData.optOut.save` (see earlier question about `do nothing` vs `do update` with respect to `updated_at`).

bchrobot (changes requested):

src/server/api/lib/message-sending.js
ajohn25: I suppose if `created_at` and `updated_at` have diff values we know that a user tried to manually opt someone out after they originally were. I don't feel strongly about keeping that info around. @hiemanshu were you thinking of anything else?

src/server/api/lib/message-sending.js
ajohn25: 1. For closing convos - I think we should change the behavior to close the convo when we get _any_ inbound message where the contact `is_opted_out` The current behavior is a little weird because if someone who's opted out responds with a regular message, we just leave that convo in `needsResponse`, but if they respond with a stop word we close the convo. 

2. External Sync - I'm probably just not familiar enough with how this is supposed to work! Can you say more about how sending over the repeated event is helpful? My thinking was that we want to notify the external system about when the contact getting opted out, not how many times they repeated that.

src/server/models/cacheable_queries/opt-out.js
ajohn25: This worked 🥳 kicked out some other dependencies too

ajohn25: But possibly broke some other things based on tests failling 😭

ajohn25: Rebasing this off #1539 based on ^

src/server/api/lib/message-sending.js
ajohn25: Both fair points! I think my only concern is being consistent with what happens "normally" (when texters are handling replies) with what auto handling is doing here. Since texters don't see opted out convos:

1. Texters currently can't opt someone back in, only a good sweeping team can from message review. @politics-rewired/organizing what message status (needsResponse/closed) do we think is most useful for sweepers for:
&nbsp; a. Stop words
&nbsp; b. Regular responses
do we need to be consistent?
2. We don't have a way to mark repeated opt out requests as a texter. I guess a sweeper could do it from message review but it's not so intuitive.

If we think that auto handling doing different things from texters isn't a problem then I think you're right on both questions

src/server/api/lib/message-sending.js
hiemanshu: > opt_out.updated_at already gets updated BEFORE UPDATE via a Postgres trigger.

If we just ignore the row, then there will be no `BEFORE UPDATE` trigger than runs to update it. 

> Question: what is the value of updating updated_at during a conflict versus just ignoring the row?

We know that the row was touched by an opt out that was triggered.

src/server/api/lib/message-sending.js
bchrobot: 1. 👍
2. I don't think there's a way to do it through the UI, but my understanding is that the question at hand is how to handle repeat inbound (e.g. `STOP`, `STOP`)?

hiemanshu: Yeah, repeated events should be fine! If we have repeated events happening, they should be passed on.

ajohn25:
> @ajohn25 can you rebase this PR off of #1539? I had assumed it was as `chore-resolve-knex` is the PR target but it doesn't look like [aae9c32](https://github.com/politics-rewired/Spoke/commit/aae9c32a433451efc36830e959010870881e7d9d) from #1539 is in `fix-duplicate-opt-outs`?

Good catch! There's those small diffs 😌

bchrobot (approved):

hiemanshu (approved):

PR #1533, opened 2022-11-17
feat: pause autosending at end of day

### LINKED ISSUE #1293: Auto-sending picks back up the following morning
**Describe the bug**

Autosending on a campaign will resume the following morning if it does not complete before texting hours close the night before. This can result in time-specific messaging being received at an incorrect later date, at which point it is no longer relevant and often confusing for the recipient.

**To Reproduce**

1.  Start auto-sending on a large campaign shortly before close
2.  Check in on the campaign at texting hours opening the next 
3.  See that autosending has resumed

**Expected behavior**

`sending` and `queued` campaigns should both automatically pause once texting hours close for the day.

**Screenshots**

N/A

**Additional context**

"when texting hours close" depends on the time zones of contacts in that campaign. One way to find this might be to select UTC offsets from all contacts based on their timezone, sort by the lowest, and use that as the cutoff.

A cron at 5 minutes past every half hour could handle performing the pauses.

hiemanshu:
## Description

Autopause auto sending campaigns at the end of the day.

End of the day is calculated based on rules here: https://rewiredcoop.slack.com/archives/C02LYSTHP0Q/p1667918970259799?thread_ts=1667845110.462929&cid=C02LYSTHP0Q

## Motivation and Context

Fixes #1293

## How Has This Been Tested?

Tested locally with mocked data. 

## Checklist:


src/server/tasks/pause-autosending.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const PAUSE_AUTOSENDING_CAMPAIGNS_TASK_IDENTIFIER =
+  "pause-autosending-campaigns";
+
+const TIMEZONES = timezones.map((tz: string) => parseIanaZone(tz));
+
+const getAutosendingCampaigns = async (): Promise<Array<CampaignRecord>> =>
+  r.knex("all_campaign").where({ autosend_status: "sending" });
+
+const getMessagesCount = async (campaignId: number) =>
```
ajohn25, line 20: Nitpick: Rename to `getQueuedMessagesCount` or similar

src/server/worker.ts
```diff
@@ -174,12 +182,26 @@ export const getScheduler = async (attempt = 0): Promise<Scheduler> => {
...
+      {
+        name: QUEUE_AUTOSEND_INITIALS_TASK_IDENTIFIER,
+        taskIdentifier: QUEUE_AUTOSEND_INITIALS_TASK_IDENTIFIER,
+        pattern: "*/1 * * * *",
+        timeZone: config.TZ
+      },
+      {
+        name: PAUSE_AUTOSENDING_CAMPAIGNS_TASK_IDENTIFIER,
+        taskIdentifier: PAUSE_AUTOSENDING_CAMPAIGNS_TASK_IDENTIFIER,
+        pattern: "10 * * * *",
```
ajohn25, line 195: Question: https://rewiredcoop.slack.com/archives/C02LYSTHP0Q/p1668811954270619?thread_ts=1667845110.462929&cid=C02LYSTHP0Q

ajohn25 (approved):
Looks right to me 🎉

src/server/tasks/pause-autosending.ts
```diff
@@ -0,0 +1,98 @@
...
+      task_identifier: retryInteractionStepIdentifier
+    })
+    .whereRaw("payload->>'campaignId' = ?", [campaignId])
+    .count("id")
+    .first();
+
+// Get Array of [tz, texting_hours_end] for whereIn query
+// DateTime end of hour takes us to the last second
+// add a minute to change time to next hour, and get the hour value
+const getHoursWithTimezones = () =>
```
bchrobot, line 33: Nitpick: match order of returned list elements

```suggestion
const getTimezonesWithHours = () =>
```

src/server/tasks/pause-autosending.ts
```diff
@@ -0,0 +1,98 @@
...
+  r
+    .knex("graphile_worker.jobs")
+    .where({
+      task_identifier: retryInteractionStepIdentifier
+    })
+    .whereRaw("payload->>'campaignId' = ?", [campaignId])
+    .count("id")
+    .first();
+
+// Get Array of [tz, texting_hours_end] for whereIn query
```
bchrobot, line 30: Nitpick: this is compared to `texting_hours_end` later on, but the values in the array are not themselves `texting_hours_end`

```suggestion
// Get Array of [tz, tz_next_hour] for whereIn query
```

src/server/tasks/pause-autosending.ts
```diff
@@ -0,0 +1,98 @@
...
+  const campaignIds = campaigns.map((c: CampaignRecord) => c.id);
+  const timeZoneParams = getHoursWithTimezones();
+
+  const contactsToUnqueueWithTimezone = await r
+    .knex("campaign_contact")
+    .join("campaign", "campaign.id", "campaign_contact.campaign_id")
+    .where({ message_status: "needsMessage" })
+    .whereIn("campaign.id", campaignIds)
+    .whereIn(
+      ["campaign_contact.timezone", "campaign.texting_hours_end"],
```
bchrobot, line 49: Required: combine these two queries with a coalesce

```suggestion
      ["coalesce(campaign_contact.timezone, campaign.timezone)", "campaign.texting_hours_end"],
```

or add where clause to match the `"campaign_contact.timezone": null,` in `contactsToUnqueueWithoutTimezone` below:

```ts
.whereNotNull("campaign_contact.timezone")
```

src/server/tasks/pause-autosending.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const unqueueAutosendingMessages: Task = async (_payload, _helpers) => {
+  const campaigns = await getAutosendingCampaigns();
+  const contacts = await getAutosendingContacts(campaigns);
+
+  await r
+    .knex("graphile_worker.jobs")
+    .where({ task_identifier: retryInteractionStepIdentifier })
+    .whereRaw("payload->>'campaignContactId' = ANY(?)", [contacts])
+    .delete();
```
bchrobot, line 79: Required: use `graphile_worker.remove_job` instead of deleting from the jobs table directly.

Manipulation of the jobs table directly is unsupported in later versions of graphile-worker so we may as well create less work for ourselves when we next update:

> The jobs table itself is not a public interface ... if you are referencing the jobs table in a database function you may have a bad time
>
> -- [v0.14.0 release notes](https://github.com/graphile/worker/blob/main/RELEASE_NOTES.md#v0140)

src/server/tasks/pause-autosending.ts
bchrobot: I would say required: `getQueuedMessageJobsCount` -- knowing the count is based on job records and not campaign_contact records is nice for quickly grokking

bchrobot (changes requested):

src/server/worker.ts
bchrobot: > wondering for pausing why do 10 mins past the hr vs. like 1 min?

Good question -- I believe that was to give currently-queued graphile-worker jobs a chance to settle before evaluating state. But with this PR, we now delete the queued jobs. I think we can actually just combine the two steps (deleting queued messages, pausing the campaign) into one task that runs on the hour

ajohn25: Got it! I think 10 mins before the hour (when the [unqueue](https://github.com/politics-rewired/Spoke/pull/1533/files#diff-eae205c288039f53fdaee668619c15cd9193fdcff31a1a632a5db5d59c9a97b8R201) currently happens) so that we're consistent with [when initials are assignable to texters](https://github.com/politics-rewired/Spoke/blob/main/schema-dump.sql#L1689)

hiemanshu:
I went back and did some more digging, and reading. With the way autosending only queues a few contacts at a time, we don’t actually need to unqueue messages, just have to make sure that there are no remaining contacts for the day, and pause the campaign.

hiemanshu:
@bchrobot @ajohn25 bunch of changes made. Second review please!

yarn.lock
```diff
@@ -18431,9 +18431,9 @@ lru-cache@^6.0.0:
     yallist "^4.0.0"
 
 luxon@^1.25.0:
-  version "1.25.0"
-  resolved "https://registry.yarnpkg.com/luxon/-/luxon-1.25.0.tgz#d86219e90bc0102c0eb299d65b2f5e95efe1fe72"
-  integrity sha512-hEgLurSH8kQRjY6i4YLey+mcKVAWXbDNlZRmM6AgWDJ1cY3atl8Ztf5wEY7VBReFbmGnwQPz7KYJblL8B2k0jQ==
+  version "1.28.0"
+  resolved "https://registry.yarnpkg.com/luxon/-/luxon-1.28.0.tgz#e7f96daad3938c06a62de0fb027115d251251fbf"
+  integrity sha512-TfTiyvZhwBYM/7QdAVDh+7dBTBA29v4ik0Ce9zda3Mnf8on1S5KJI8P2jKFZ8+5C0jhmr0KwJEO/Wdpm0VeWJQ==
```
bchrobot, line 18436: Required: revert this change or add corresponding change to `package.json` with a note in PR description about relevance to PR

src/server/tasks/pause-autosending.ts
```diff
@@ -27,72 +24,66 @@ const getMessagesCount = async (campaignId: number) =>
...
   for (const campaign of runningCampaigns) {
-    const { count: messagesToSend } = await getMessagesCount(campaign.id);
+    const { count: messagesToSendCount } = await getQueuedMessagesJobsCount(
+      campaign.id
+    );
+
+    const elligibleContactsCount = await getElligibleContactsCount(campaign);
 
-    if (messagesToSend === 0) {
+    if (Number(messagesToSendCount) + Number(elligibleContactsCount) === 0) {
```
bchrobot, line 80: Nitpick: converting `elligibleContactsCount` to `number` seems unnecessary as `getElligibleContactsCount()` already does its own type conversion (line 65). Maybe use `parseInt()` instead for better TS type agreement?

src/server/tasks/pause-autosending.ts
```diff
@@ -27,72 +24,66 @@ const getMessagesCount = async (campaignId: number) =>
     .count("id")
     .first();
 
-// Get Array of [tz, texting_hours_end] for whereIn query
+// Get Array of [tz, tz_next_hour] for whereIn query
```
bchrobot, line 27: Required: s/tz_next_hour/tz_current_hour to reflect the switch to `startOf("hour")`

src/server/tasks/pause-autosending.ts
```diff
@@ -27,72 +24,66 @@ const getMessagesCount = async (campaignId: number) =>
...
-  const contactsToUnqueueWithTimezone = await r
+  const [{ count: contactsWithTimezoneCount }] = await r
     .knex("campaign_contact")
     .join("campaign", "campaign.id", "campaign_contact.campaign_id")
-    .where({ message_status: "needsMessage" })
-    .whereIn("campaign.id", campaignIds)
-    .whereIn(
+    .where({ message_status: "needsMessage", campaign_id: campaign.id })
+    .whereNotNull("campaign_contact.timezone")
+    .whereNotIn(
```
bchrobot, line 44: Question: My understanding is that using `whereNotIn()` will only work if all contacts' are in the same time zone. As written, this only excludes contacts whose texting hours _just_ ended at the previous hour mark, meaning it will include a contact whose texting hours closed _more_ than an hour ago (and should still be considered untextable). Does that seem right?

Would it be easier to push this down to the DB using a where clause similar to the `assignable_needs_message ` view?

https://github.com/politics-rewired/Spoke/blob/5d073c07e0c890cf02f8609dba4393fc22a774c0/schema-dump.sql#L1689

bchrobot (changes requested):

src/server/tasks/pause-autosending.ts
hiemanshu: Fixed in the next commit.

yarn.lock
hiemanshu: This is a required change. It's needed for `DateTime.now()` to work correctly. I'll update the package.json

src/server/tasks/pause-autosending.ts
hiemanshu: Good question. Yeah seems like it does that. I tried to test with a lot of different scenarios but didn't catch that. I'm going to use the same check that's used by queueing initials job, so it's on parity with that.

bchrobot (approved):

PR #1540, opened 2022-11-30
feat!: allow uploading arbitrary csvs

### LINKED ISSUE #1156: Upload arbitrary CSVs and map columns on the client
Spoke is used in many different ways by many different organizations. Before submitting a feature request please think about whether it addresses a use case specific to your organization or whether it will provide value across many different use cases.

If you're unsure, drop us a line at support@spokerewired.com and we would be more than happy to talk through it with you!

Importing a CSV into Spoke requires manual or scripted cleanup to shape the CSV columns into the expected `firstName`, `lastName`, `cell`, `zip` columns.

Similar to how Flatfile.io manages CSV field mapping.

A feature at upload time that:
- parses the CSV client-side
- confirms with the user that the first row contains headers (with preview)
- asks the user to match the First Name, Last Name, Cell Phone, and ZIP inputs to the appropriate columns in their CSV
- sends this mapping with the CSV payload to the backend GraphQL API
- uses the mapping to pick columns from the CSV and produce a properly formatted structure for further processing

As a Spoke customer on a campaign, I wrote a tool for organizers so they could upload CSVs, match columns, and autogenerate CSVs with the header columns that Spoke was expecting. They would then save this CSV and upload it into Spoke. We could make the process simpler and more accessible to all Spoke users, not just tech-savvy ones able to write apps or scripts, by integrating this functionality into Spoke.

**Additional context**

Screenshots from the campaign tool:

Upload a CSV
image

Map columns
image

Download file
image

hiemanshu:
## Description

Allow uploading arbitrary CSVs and allowing column remapping 

## Motivation and Context

Fixes #1156 

## How Has This Been Tested?

Tested locally with different variants of CSVs

Screenshot 2022-11-30 at 1 47 15 PM
Screenshot 2022-11-30 at 1 47 28 PM
Screenshot 2022-11-30 at 1 47 55 PM
Screenshot 2022-11-30 at 1 48 16 PM

## Checklist:

---
  - https://app.asana.com/0/0/1203491785595177

### DIFF
--- a/migrations/20221125080945_add_column_mapping_to_campaign.js
+++ b/migrations/20221125080945_add_column_mapping_to_campaign.js
@@ -36,8 +36,8 @@ exports.up = function up(knex) {
           is_approved,
           autosend_status,
           autosend_user_id,
-          autosend_limit,
           messaging_service_sid,
+          autosend_limit,
           column_mapping
         from all_campaign
         where is_template = false;
@@ -81,8 +81,8 @@ exports.down = function down(knex) {
           is_approved,
           autosend_status,
           autosend_user_id,
-          autosend_limit,
-          messaging_service_sid
+          messaging_service_sid,
+          autosend_limit
         from all_campaign
         where is_template = false;
 
--- a/migrations/20221125080945_add_column_mapping_to_campaign.js
+++ b/migrations/20221125080945_add_column_mapping_to_campaign.js
@@ -1,0 +1,299 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function up(knex) {
+  return knex.schema
+    .alterTable("all_campaign", (table) => {
+      table.text("column_mapping").nullable();
+    })
+    .then(() =>
+      knex.schema.raw(`
+        create or replace view campaign as
+        select
+          id,
+          organization_id,
+          title,
+          description,
+          is_started,
+          due_by,
+          created_at,
+          is_archived,
+          use_dynamic_assignment,
+          logo_image_url,
+          intro_html,
+          primary_color,
+          texting_hours_start,
+          texting_hours_end,
+          timezone,
+          creator_id,
+          is_autoassign_enabled,
+          limit_assignment_to_teams,
+          updated_at,
+          replies_stale_after_minutes,
+          landlines_filtered,
+          external_system_id,
+          is_approved,
+          autosend_status,
+          autosend_user_id,
+          autosend_limit,
+          messaging_service_sid,
+          column_mapping
+        from all_campaign
+        where is_template = false;
+      `)
+    );
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function down(knex) {
+  return knex.schema
+    .raw(
+      `
+      drop view campaign cascade;
+      create view campaign as
+        select
+          id,
+          organization_id,
+          title,
+          description,
+          is_started,
+          due_by,
+          created_at,
+          is_archived,
+          use_dynamic_assignment,
+          logo_image_url,
+          intro_html,
+          primary_color,
+          texting_hours_start,
+          texting_hours_end,
+          timezone,
+          creator_id,
+          is_autoassign_enabled,
+          limit_assignment_to_teams,
+          updated_at,
+          replies_stale_after_minutes,
+          landlines_filtered,
+          external_system_id,
+          is_approved,
+          autosend_status,
+          autosend_user_id,
+          autosend_limit,
+          messaging_service_sid
+        from all_campaign
+        where is_template = false;
+
+      create or replace view assignable_campaigns as (
+        select id, title, organization_id, limit_assignment_to_teams
+        from campaign
+        where is_started = true
+          and is_archived = false
+          and is_autoassign_enabled = true
+      );
+
+      create or replace view assignable_campaign_contacts as (
+        select
+          campaign_contact.id, campaign_contact.campaign_id,
+          campaign_contact.message_status, campaign.texting_hours_end,
+          campaign_contact.timezone::text as contact_timezone
+        from campaign_contact
+        join campaign on campaign_contact.campaign_id = campaign.id
+        where assignment_id is null
+          and is_opted_out = false
+          and archived = false
+          and not exists (
+            select 1
+            from campaign_contact_tag
+            join tag on campaign_contact_tag.tag_id = tag.id
+            where tag.is_assignable = false
+              and campaign_contact_tag.campaign_contact_id = campaign_contact.id
+          )
+      );
+
+      create or replace view assignable_needs_message as (
+        select acc.id, acc.campaign_id, acc.message_status
+        from assignable_campaign_contacts as acc
+        join campaign on campaign.id = acc.campaign_id
+        where message_status = 'needsMessage'
+          and (
+              ( acc.contact_timezone is null
+                and extract(hour from CURRENT_TIMESTAMP at time zone campaign.timezone) < campaign.texting_hours_end
+                and extract(hour from CURRENT_TIMESTAMP at time zone campaign.timezone) >= campaign.texting_hours_start
+              )
+            or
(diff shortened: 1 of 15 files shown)


hiemanshu:
@politics-rewired/organizing copy needed as always for the popup dialog.

ninabaldwin:
@hiemanshu We have some ideas to change it so that it is required to map every time someone loads a CSV. I have uploaded a video on how actionnetwork does it and we pretty much want to have it do the same thing:

https://user-images.githubusercontent.com/98056070/205129515-dcbf208e-7db9-47cb-be56-576dee6c6e12.mp4

On top of that, we also want it to automatically map first name, last name, etc, any field that we already map. For example, if the CSV has the headers firstName, lastName and cell then when the mapping pops up it automatically fills those in as First Name, Last Name, and Cell Phone Number. We want that for any header we accept today (i.e. firstname, firstName etc, I don't know everything we accept today) and nothing else. Everything else would be loaded with the name provided (defaults as "Load as variable"). 

So we would only "map" the special fields we use and everything else would be loaded as they have named in the file. I would recommend using normal names from everything moving forward when selecting the "mapped" field from the drop-down (not camel case) >

First Name
Last Name
Cell Phone Number
Zip
External ID
(I think these are only special fields that require mapping? not 100% sure)

And maybe everything else would default to "Load as variable"

An example:
Load a CSV with headers: lastname, phone, address, zip code, firstName

Mapping would default to:

lastname > Last Name
phone > Load as variable
address > Load as variable
zip code > Load as variable
firstName > First Name

They would need to update phone and zip code to "Cell Phone Number" and "Zip" from a drop down that only contains the last special fields that can be mapped.


Ok clearly this is a lot of info but let me know if there is anything else I can answer!

ajohn25 (changes requested):

migrations/20221125080945_add_column_mapping_to_campaign.js
```diff
@@ -0,0 +1,299 @@
...
+          limit_assignment_to_teams,
+          updated_at,
+          replies_stale_after_minutes,
+          landlines_filtered,
+          external_system_id,
+          is_approved,
+          autosend_status,
+          autosend_user_id,
+          autosend_limit,
+          messaging_service_sid
```
ajohn25, line 85: Required: Same as above
```suggestion
          messaging_service_sid,
          autosend_limit
```

hiemanshu:
> So we would only "map" the special fields we use and everything else would be loaded as they have named in the file. I would recommend using normal names from everything moving forward when selecting the "mapped" field from the drop-down (not camel case) 

So we won't allow custom column remappings of any kind? The sense I got from conversations was that we wanted to allow that to happen.

> On top of that, we also want it to automatically map first name, last name, etc, any field that we already map. For example, if the CSV has the headers firstName, lastName and cell then when the mapping pops up it automatically fills those in as First Name, Last Name, and Cell Phone Number.

It already does this. We can make a more exhaustive list if we need, here's what it currently is: https://github.com/politics-rewired/Spoke/blob/563dc7a64b3f0e1088965de3ab41abdf12ee0e85/src/containers/AdminCampaignEdit/sections/CampaignContactsForm/components/ConfigureColumnMappingDialog.tsx#L17-L28

It takes those names, and will automatically pop in values.

ajohn25:
@hiemanshu you can ignore the section about changing what gets mapped since we're already doing it. We should change the dialog to pop up every time someone loads a CSV, so that it's required to do the mapping when you upload.

hiemanshu:
> @hiemanshu you can ignore the section about changing what gets mapped since we're already doing it. We should change the dialog to pop up every time someone loads a CSV (instead of having to click the Configure button). We'd like it to be required for the user to correct/confirm the mapping when you upload, and that should be the only change required from the first organizing comment!

So currently the user is required to correct/confirm the mapping when they upload. You can't save the contacts without a correct mapping. I'll make the dialog open on loading the CSV, looks like the only change since other changes already exist.

@politics-rewired/organizing do you think it would be helpful to have staging run this feature branch so that y'all can test the current behavior and see how that works?

ninabaldwin:
@hiemanshu yes that would be great!

ajohn25 (approved):

migrations/20221125080945_add_column_mapping_to_campaign.js
```diff
@@ -0,0 +1,299 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function up(knex) {
+  return knex.schema
+    .alterTable("all_campaign", (table) => {
```
bchrobot, line 7: Thinking about future work in this area, I think adding a new table (something like `campaign_contact_upload`), rather than adding another column to the ever-widening `campaign` table, is the way to go. My apologies for not making the connections earlier, before you started working on this 😞. I don't think it will require any changes to business logic though, just bits around data plumbing.

A few of those future work items/benefits of a separate table are:

An idea that has been discussed a few times, but not turned into a task, is moving contact CSV upload to a cloud storage-backed job. When the user selects a local CSV, Spoke would request a signed upload URL from the backend and then upload the file there. When the contacts section is saved, an async task would be created that would download the file from cloud storage and process it as we currently do. This makes contact upload more robust to mid-upload pod restarts, gets rid of shoving an entire CSV into `job_request.payload`, and makes the user experience snappier as the CSV no longer needs to be processed synchronously during `editCampaign`. #1434 already added support for importing contacts from URL, including mapping column names.

We know we will want to store more things about the source CSV file for #1144.

Upload stats are currently haphazardly pulled from the plain text `job_request.result` column, but the format and structure of what is saved to `result` has changed over time. With a dedicated table for CSV uploads, those stats could be stored in a structured way.

No changes to the various campaign views!

hiemanshu:
@politics-rewired/organizing https://stagingenv.spokerewired.com has this feature available now!

ninabaldwin:
@hiemanshu we want some changes!

- Change to a drop-down menu (with options "First Name," "Last Name," "Cell Phone," "External ID," "Zip" and "Create New Field")
- Do not use wording "remap to" and instead just say "Create field with name as is"
- Create new field allows them to change the name of the header to whatever they want (include an error that doesn't allow duplicates- for example it won't allow First Name

src/server/api/lib/campaign.ts
```diff
@@ -471,9 +472,26 @@ export const editCampaign = async (
...
+        DO UPDATE SET column_mapping = EXCLUDED.column_mapping, updated_at = CURRENT_TIMESTAMP 
+        RETURNING *;
+      `,
+      [
+        r.knex("campaign_contact_upload").insert({
+          campaign_id: id,
+          column_mapping: JSON.stringify(columnMapping)
+        })
+      ]
+    );
```
bchrobot, line 494: Suggestion: just use a raw query here. The composition makes this harder to grok.

The update to `updated_at` also duplicates work the `_500_campaign_contact_upload_updated_at` trigger is doing.

ajohn25:
A few more changes we would like (hopefully last round 😭 :
- In addition to the duplicate error, the drop downs should have a field removed from the list if it's already been used. Ex. if the first row has "First Name" selected, no other row has "First Name" in the drop down until it's deselected in the first row.
- For columns that are not in the drop down, the autocomplete box should have a gray placeholder with "Create as" + the name of the column in the csv. Ex. an ask column by default says `Create as "ask"`
- We'd like to add a "Type new column name" option in the drop down. This is just to let users know they're able to type in their own name, since that may not be obvious at a glance. When you click that option, we just hide the drop down and let the user type.

hiemanshu:
@politics-rewired/organizing changes are up on staging for 1 and 2. Last change needed to finish up this PR is having copy for the document on how to use this, and what is required, etc. Let me know when you have the copy ready, and I'll push it to staging, so you can have the document prepared too.

ninabaldwin:
@hiemanshu you are saying we need to update this article: https://docs.spokerewired.com/article/64-contact-lists and any other articles that mention how to upload a csv?

hiemanshu:
> @hiemanshu you are saying we need to update this article: https://docs.spokerewired.com/article/64-contact-lists and any other articles that mention how to upload a csv?

Yeah. And also some copy for the dialog, talking about the required fields, and additional fields, and that you can type in custom fields too. Feel free to add in / remove any information that would make more sense going into the document.

PR #1574, opened 2023-02-27
chore: replace useless map with copy-on-destructure

bchrobot:
## Description

This replaces a useless `map()` with a useful copy via array destructuring (to avoid accidentally mutating source array).

## Motivation and Context

Comment in #1573.

See https://medium.com/@abdulrasheed.formal/different-ways-to-clone-copy-an-array-in-javascript-be0f5a2ec24e#7963 for more info about destructuring/spreading arrays to copy them.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:


ajohn25 (approved):
Suggestion: Something not mentioned on the linked article is[ `Array.from()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) which is maybe more expressive for "I'm making a copy here"?

PR #1576, opened 2023-02-27
chore: use more accurate variable naming

bchrobot:
## Description

This rename makes it clear that the objects being passed around are CampaignContactTags and not Tags.

## Motivation and Context

Without type checking it is easy to make a mistake if variable names are misleading.

## How Has This Been Tested?

This has been tested locally.

N/A

## Documentation Changes

N/A

## Checklist:

### DIFF
--- a/src/containers/AssignmentTexterContact/index.jsx
+++ b/src/containers/AssignmentTexterContact/index.jsx
@@ -474,8 +474,8 @@ export class AssignmentTexterContact extends React.Component {
     }
   };
 
-  handleApplyTagsAndMoveOn = (addedTags, removedTags) => {
-    this.handleApplyTags(addedTags, removedTags, async () => {
+  handleApplyTagsAndMoveOn = (addedContactTags, removedContactTags) => {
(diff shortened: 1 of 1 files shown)


src/containers/AssignmentTexterContact/index.jsx
```diff
@@ -470,23 +470,27 @@ export class AssignmentTexterContact extends React.Component {
...
     });
 
-    this.tagContact(addedTags, removedTags);
+    this.tagContact(addedContactTags, removedContactTags);
 
     if (callback) {
       this.setState(
-        { addedTags, removedTags, isTagEditorOpen: false },
+        {
+          addedTags: addedContactTags,
```
ajohn25, line 478: Question: Should we rename the state variables too to make this clear across the board?

ajohn25 (approved):

src/containers/AssignmentTexterContact/index.jsx
bchrobot: That was more than I wanted to touch without a TypeScript rewrite and a lot more time manually testing. But, yeah, that would be good to do longer term.

PR #1610, opened 2023-05-15
feat(campaign-interaction-steps): add save button to top of section

### LINKED ISSUE #1547: Add Save button to top of Interaction Steps section when unfurled
When making an edit to a script with many interaction steps, a user has to navigate all the way to the bottom of the Interactions Step section to hit Save. For users who don't use keyboard shortcuts or who are just editing the initial message, it may take less time to navigate to the top of the section than the bottom.  

The same Save button that appears at the top of the Interaction Steps section should appear at the top right side of the section, right-justified but aligned with Copy Block and Paste Block buttons.

Saving upon edit; without edit history and the ability to restore previous versions, this is not desired behavior.

henryk1229:
## Description

This pr adds the save button at the bottom of the interaction steps section of a campaign to the top of the section as well, so that users can scroll up to save when editing campaigns with long scripts

## Motivation and Context

It closes #1547 

## How Has This Been Tested?

Locally

Screenshot 2023-05-15 at 12 40 52 PM

## Documentation Changes

n/a

## Checklist:

### DIFF
--- a/src/containers/AdminCampaignEdit/sections/CampaignInteractionStepsForm/index.tsx
+++ b/src/containers/AdminCampaignEdit/sections/CampaignInteractionStepsForm/index.tsx
@@ -389,9 +389,22 @@ const CampaignInteractionStepsForm: React.FC<InnerProps> = (props) => {
         title="What do you want to discuss?"
         subtitle="You can add scripts and questions and your texters can indicate responses from your contacts. For example, you might want to collect RSVPs to an event or find out whether to follow up about a different volunteer activity."
       />
-      <Box m={2}>
-        <ScriptPreviewButton campaignId={campaignId} />
-      </Box>
+      <div style={{ display: "flex", justifyContent: "space-between" }}>
+        <Box m={2}>
+          <ScriptPreviewButton campaignId={campaignId} />
+        </Box>
+        <Box m={2}>
+          <Button
+            {...dataTest("interactionSubmit", true)}
+            variant="contained"
+            color="primary"
+            disabled={isSaveDisabled}
+            onClick={handleSave}
+          >
+            {finalSaveLabel}
+          </Button>
+        </Box>
+      </div>
       {renderInvalidScriptFields()}
       <InteractionStepCard
         interactionStep={finalFree}


src/containers/AdminCampaignEdit/sections/CampaignInteractionStepsForm/index.tsx
```diff
@@ -389,9 +389,22 @@ const CampaignInteractionStepsForm: React.FC<InnerProps> = (props) => {
         title="What do you want to discuss?"
         subtitle="You can add scripts and questions and your texters can indicate responses from your contacts. For example, you might want to collect RSVPs to an event or find out whether to follow up about a different volunteer activity."
       />
-      <Box m={2}>
-        <ScriptPreviewButton campaignId={campaignId} />
-      </Box>
+      <div style={{ display: "flex", justifyContent: "space-between" }}>
```
henryk1229, line 392: I looked through the onboarding and contributing docs but didn't see any guidance around css/styling, so I kept this as an inline change for simplicity's sake. I'm happy to change it if there's a convention I should be following, lmk!

lediur:
Maybe this is a follow-up issue / PR, but I wonder if this Save button should just save, rather than "Save and goto next section", since the section collapse toggle button is right there.

If we decide that's a good idea, I would rename the button to "Save interactions"

src/containers/AdminCampaignEdit/sections/CampaignInteractionStepsForm/index.tsx
ajohn25: Yeah agree on no real convention. I've followed a rule of thumb of if it goes over one line - pull it out of the body. Curious if @hiemanshu has thoughts around this too.

ajohn25: Suggestion: See if you can use a [`Grid`](https://v4.mui.com/components/grid/#grid) with Grid items to get the same outcome. [See example here](https://github.com/politics-rewired/Spoke/blob/main/src/containers/AdminCampaignStats/components/DeliverabilityStats.tsx)  Inline CSS for the div with `Box` containers for the buttons feels mismatched I think.

ajohn25:
> Maybe this is a follow-up issue / PR, but I wonder if this Save button should just save, rather than "Save and goto next section", since the section collapse toggle button is right there.
> 
> If we decide that's a good idea, I would rename the button to "Save interactions"

agree with this suggestion and think it can happen in this PR!

ajohn25 (approved):

PR #1617, opened 2023-05-24
fix: cannot save assign control max count

mkoontz-rewired:
## Description

Fixing an issue that showed up in helpscout and fixing other small issues in the same area. See commit messages for more detail.

## Motivation and Context

- Root cause fix for https://secure.helpscout.net/conversation/2252793441/8846/ https://rewiredcoop.slack.com/archives/C02L56CP92A/p1684939561227669
- Also realized there's no way to clear escalation tags while fixing this, so fixed that, too. 
- Some small styling changes, see screenshot

## How Has This Been Tested?

locally

OLD:
Screen Shot 2023-05-24 at 2 49 24 PM

NEW:
Screen Shot 2023-05-24 at 2 43 44 PM

## Checklist:

### DIFF
--- a/src/components/QuickSearchToolbar.tsx
+++ b/src/components/QuickSearchToolbar.tsx
@@ -6,7 +6,7 @@ import React from "react";
 
 interface QuickSearchToolbarProps {
   value: string;
-  onChange(): void;
+  onChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement>;
   clearSearch(): void;
 }
 
--- a/src/containers/AdminAssignmentControl/components/AssignmentRow.tsx
+++ b/src/containers/AdminAssignmentControl/components/AssignmentRow.tsx
@@ -67,7 +67,13 @@ const AssignmentRow: React.FC<Props> = (props) => {
     isRowDisabled || !isAssignmentEnabled || id === "general";
 
   return (
-    <Grid container spacing={2} wrap="nowrap" alignItems="center">
+    <Grid
+      container
+      spacing={2}
+      wrap="nowrap"
+      alignItems="center"
+      justifyContent="space-between"
+    >
       <Grid item>
         <Chip label={title} style={{ color: textColor, backgroundColor }} />
       </Grid>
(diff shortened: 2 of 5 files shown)


src/components/QuickSearchToolbar.tsx
```diff
@@ -6,7 +6,7 @@ import React from "react";
 
 interface QuickSearchToolbarProps {
   value: string;
-  onChange(): void;
+  onChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement>;
```
hiemanshu, line 9: Question: Is this a related change? If not, should be removed.

src/containers/AdminAssignmentControl/components/AssignmentRow.tsx
```diff
@@ -37,17 +41,18 @@ const AssignmentRow: React.FC<Props> = (props) => {
...
     onChange({
       maxRequestCount: maxRequestCount ? parseInt(maxRequestCount, 10) : null
     });
+  };
 
   const handleChangeEscalationTags = (
     _event: React.ChangeEvent<any>,
-    value: Array<TagWithTitle> | null
-  ) => {
+    value: TagWithTitle[]
```
hiemanshu, line 54: nit: I prefer the `Array<type>` over `Type[]` style, reads better. Don't have to make that change, but we should settle on a convention 😓

hiemanshu (approved):
LGTM

src/containers/AdminAssignmentControl/components/AssignmentRow.tsx
ajohn25: i like `Type[]` lol we should vote

src/components/QuickSearchToolbar.tsx
mkoontz-rewired: It's related to me. The max count bug was caused by switching to @material-ui/core/TextField from material-ui/TextField, which changed the typing of onChange. 

I was looking through other instances of onChange used with <TextField> and saw that this one had bad typing. Feels related enough to the core issue involved.

src/containers/AdminAssignmentControl/components/AssignmentRow.tsx
mkoontz-rewired: yeah fine with either but I also like Type[] lol

ajohn25 (approved):

PR #1620, opened 2023-06-08
chore: add july copy updates

ajohn25:
## Description
This PR prepares copy changes which should be released to Spoke instances on 7/1/23

## Motivation and Context
#ops-2023-spoke-pricing project

## How Has This Been Tested?

N/A

## Checklist:

### DIFF
--- a/src/containers/AdminCampaignEdit/sections/CampaignFilterLandlinesForm.tsx
+++ b/src/containers/AdminCampaignEdit/sections/CampaignFilterLandlinesForm.tsx
@@ -93,10 +93,9 @@ class FilterLandlinesForm extends React.Component<Props, State> {
             !landlinesFiltered && (
               <span>
                 <p>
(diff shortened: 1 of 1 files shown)


ninabaldwin:
What about "for the lowest text message prices and best deliverability!" 
and
"but you will be charged at our higher unregistered prices."

I was also just thinking we could just say that 10DLC registered is the lowest price and focus on a more of "you can save money if you do 10DLC" versus "you lose money if you send unregistered" strategy

ninabaldwin:
From Kohlee:
I'd lose the comma in first image between "charges for messaging" and "and may see lower message deliverability"
For second, maybe
"Filtering your contact list to remove landlines or otherwise untextable numbers takes just a few minutes and helps you avoid paying for messages to numbers that won't receive them.
This service is optional and free of charge, but if you're pretty sure your phone numbers are valid, you can skip this section."

PR #1629, opened 2023-06-28
feat: indicate new campaign id(s) from template

### LINKED ISSUE #1563: Indicate new campaign ID when creating from template
When creating from a template, the user isn't made aware of the new campaign ID. To navigate to the newly created campaign, they need to navigate to the Campaigns page and scroll to the bottom. For many organizations, this list of unarchived campaigns can be long. 

A popup should appear at the bottom of the screen to indicate the new campaign ID. This behavior could mirror what happens when "Copy Campaign" is selected from a Campaign Details page.

A dialog box could offer the user to open the newly-created campaign's edit page upon clicking "Create from template"
**Additional context**

mkoontz-rewired:
## Description

Added a snackbar that opens up after creating campaign(s) from the template. Snackbar content includes links to open up new windows to the created campaigns.

## Motivation and Context

Closes #1563 

## How Has This Been Tested?

locally

## Scre
Screenshot 2023-06-28 at 12 31 17 PM
enshots (if appropriate):

## Checklist:


src/components/CreateCampaignFromTemplateDialog.tsx
```diff
@@ -6,39 +7,58 @@ import DialogContentText from "@material-ui/core/DialogContentText";
...
   const { data, error } = useGetTemplateCampaignsQuery({
     variables: { organizationId: props.organizationId }
   });
   const [
     createFromTemplate,
     { loading: working }
   ] = useCreateCampaignFromTemplateMutation({
-    refetchQueries: [GetAdminCampaignsDocument]
+    refetchQueries: [GetAdminCampaignsDocument],
+    onCompleted: (onCompletedData) =>
```
ajohn25, line 50: Suggestion: Add the state changes from `onClose()` to the completion handler in the parent, then just [call `onCreateTemplateCompleted()` instead of  `onClose()` after the mutation returns](https://github.com/politics-rewired/Spoke/pull/1629/files#diff-42c2da86c4e0d4ca05a788365c3695853e2a9e9651a033932219aac1436ddc78R103)

ajohn25 (approved):

hiemanshu (approved):

src/containers/AdminCampaignList.jsx
```diff
@@ -293,9 +328,46 @@ class AdminCampaignList extends React.Component {
         ) : null}
         <CreateCampaignFromTemplateDialog
           organizationId={organizationId}
-          open={this.state.createFromTemplateOpen}
+          open={createFromTemplateOpen}
           onClose={() => this.setState({ createFromTemplateOpen: false })}
+          onCreateTemplateCompleted={this.onCreateTemplateCompleted}
```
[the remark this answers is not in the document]
mkoontz-rewired: happy to change to this.handleCompleteCreateTemplate, but wondering why it's preferred to have the prop name be onComplete rather than onCreateTemplateComplete? I like more verbose props as it's _possible_ whatever child component can complete multiple different types of actions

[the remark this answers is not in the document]
mkoontz-rewired: Actually, I feel like this `this.handleCreateTemplateCompleted` is more accurate than this version `this.handleCompleteCreateTemplate`

The first implies we're calling some side effects after "create template (mutation) has been completed" whereas the second suggests we're taking some action to complete "create template".

Does that make sense? I know this is a nit, but if it's an unspoken convention I'd rather talk about it.
So actually I'd prefer, `onCreateTemplateCompleted=this.handleCreateTemplateCompleted`

src/components/CreateCampaignFromTemplateDialog.tsx
mkoontz-rewired: I know this is a suggestion, but I want to engage to make sure I know why I'm making certain decisions.

1. By using `onCompleted`, I don't have to do the error checking/data checking myself in `handleClickCreate`. I can guarantee that `handleCreateTemplateCompleted` will be called when we have data and no errors without needing to recreate that logic. I can see an argument about keeping mutation-related logic within `handleClickCreate` in one place. If that's the argument you have, I can move the callback to handleClickCreate and add the data/error check.

2. If I combine onClose to the completion handler `handleCreateTemplateCompleted`, then the completion handler is no longer a pure function. It would become a fn that does "create template completed" state stuff _and_  have unexpected close behavior. This additional behavior wouldn't be obvious from the fn name alone when you're within the context of the CreateCampaignFromTemplateDialog component, which to me is a sign of a badly designed fn.

src/components/CreateCampaignFromTemplateDialog.tsx
```diff
@@ -6,39 +7,58 @@ import DialogContentText from "@material-ui/core/DialogContentText";
...
   useGetTemplateCampaignsQuery
 } from "@spoke/spoke-codegen";
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 
 export interface CreateCampaignFromTemplateDialogProps {
   organizationId: string;
   open: boolean;
+  onCreateTemplateCompleted: (
+    data: MutationResult<CreateCampaignFromTemplateMutation>["data"],
```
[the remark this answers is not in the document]
mkoontz-rewired: done, thx! 🙏

src/containers/AdminCampaignList.jsx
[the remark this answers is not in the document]
ajohn25: im good with all your thoughts here, thanks for walking me through those! also cc @hiemanshu if you have any additional thoughts cause i've generally been following your (virtual) lead here

src/components/CreateCampaignFromTemplateDialog.tsx
ajohn25: > I can see an argument about keeping mutation-related logic within handleClickCreate in one place. If that's the argument you have, I can move the callback to handleClickCreate and add the data/error check.

this is what i was thinking - i do (personally) prefer having the logic in one place when it's reasonable to

> It would become a fn that does "create template completed" state stuff and have unexpected close behavior. 

feel free to not dive into this 😆 but would you say similar about the `handleDelete` handler here? https://github.com/politics-rewired/Spoke/blob/main/src/containers/AdminTemplateCampaigns/components/TemplateCampaignRow.tsx#L114

mkoontz-rewired: Ok! I can move the callback logic into handleClickCreate 👍 

I think handleDelete is less of a problem here because every fn called within `handleDelete` is within the same component. But if you wanted to be super clear `handleDelete` should be renamed to `handleDeleteTemplateButtonClick`. The more precise name shows that it does various things onClick of the delete template button rather than being a fn that should only handle deletion of "something".

mkoontz-rewired: idk if I'm being clear. Precise naming makes it easier to navigate components and switch out the particulars of implementation when refactoring.

ajohn25: yup totally makes sense 👍🏾

src/containers/AdminCampaignList.jsx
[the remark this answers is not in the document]
hiemanshu: Most libraries we've been using, and the "unspoken" convention has been to have props be small `onComplete`, `onClose`, `onOpen`, etc. And the function that handles it being more verbose `handleClickCreateNewCampaign` or similar. It keeps things in line with the current way we've been doing things, and these were done using something tips from Material UI components using similar conventions.

PR #1654, opened 2023-08-04
fix(bulk script editor): restrict to user organization

### LINKED ISSUE #1640: Bulk script editor is applied across the instance instead of limited to the organization
**Describe the bug**
Previously the bulk script editor would apply only to messages within an organization. We received a report that it is now being applied to messages across the entire instance. This can be problematic as administrators in one organization can make changes to other organizations without their knowledge.

We need to revert the behavior back to applying only to messages within that organization.

**To Reproduce**

From within an instance that has multiple organizations, find a phrase that occurs in a text message script in both organizations.

1. Go to "Bulk Script Editor"
2. In replace text, type the phrase you found
3. Type in text that it will be replaced with
4. Click find and replace
5. Review scripts with the phrase to confirm that it was changed in different organizations.

**Expected behavior**
Only phrases within that organization are updated. Other organizations should have no change.

ajohn25:
## Description

This restricts the bulk script editor to only apply to the user's organization

## Motivation and Context

Closes #1640

## How Has This Been Tested?

This has been tested locally

## Checklist:

### DIFF
--- a/libs/gql-schema/schema.ts
+++ b/libs/gql-schema/schema.ts
@@ -48,6 +48,7 @@ const rootSchema = `
     searchString: String!
     replaceString: String!
     campaignIds: [String!]!
+    organizationId: String!
   }
 
   input ContactActionInput {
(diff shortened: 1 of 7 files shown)


henryk1229 (approved):
lgtm!

hiemanshu (approved):

Issue #1526, opened 2022-11-08
Automatic Reply Handling


ajohn25:
# Overview
Allow users to set up campaigns so that common responses to an initial message can be automatically handled.

# Context

For some organizations, many replies that are handled by texters currently do not actually require much thought, and can be automatically parsed and handled by Spoke automatically.

# Goals and Non-Goals
Goals:
- Allow users to specify what "common responses" are to an initial message and what survey responses should be sent automatically as the reply. In other words, allow users to specify what messages Spoke should automatically respond to and how.
- Refactor [`retry-interaction-step`](https://github.com/politics-rewired/Spoke/blob/main/src/server/tasks/retry-interaction-step.ts) to allow for sending `script_options` that are not the root step.

Non-Goals:
- Handling responses after the initial reply (The flow of a conversation can change significantly after the first response and make auto handling difficult, but @politics-rewired/organizing should also weigh in if we can think of a good way to do this)

# Milestones

This will be completed in a single PR.

# Existing Solution

Users handle all replies manually.

# Proposed Solution

# Alternative Solutions

Predetermine a common list of response words as we've done in #1110 Since we will have to store the pairs of keywords and interaction steps to send in response in the database anyway, predetermining a common list does not save much work overall, and limits more flexible use of this feature.

# Testability, Monitoring and Alerting

Write test cases as part of this PR.

# Cross-Team Impact

Will require documentation updates.

# Open Questions

# Detailed Scoping and Timeline

ajohn25:
@hiemanshu This is a lot smaller in scope than #1516 so I'm curious whether you think this one should wait for that to be completed in terms of front end changes.

bchrobot:
@ajohn25 a few questions to confirm understanding and a few suggestions:

Question: the overall approach being proposed is that response keywords would be associated with a _child_ interaction step such that inbound messages in response to the _parent_ interaction step's question would a) automatically mark that question response and b) send the next script automatically (if a script exists for the child step), is that right?

> Refactor [retry-interaction-step](https://github.com/politics-rewired/Spoke/blob/main/src/server/tasks/retry-interaction-step.ts) and [message-sending.js](https://github.com/politics-rewired/Spoke/blob/main/src/server/api/lib/message-sending.js) to allow for sending script_options that are not the root step.

Question: the goal here is to support automated sending of non-root interaction steps, is that right?

> Add a `response_words text[]` column to `interaction_step` to store the words to respond to for each interaction step.

[...]

PR #1663, opened 2023-08-12
feat: support auto reply handling

ajohn25:
## Description

This PR adds all backend changes required to support auto reply handling.

This does _not_ expose any auto reply functionality to users intentionally, to allow for an easier slow rollout of this feature if desired.

## Motivation and Context

Part 1 of 2 for #1526 

## How Has This Been Tested?

See 

## Checklist:

### DIFF
--- a/migrations/20230811182109_auto-replies-autoassignment.js
+++ b/migrations/20230811182109_auto-replies-autoassignment.js
@@ -1,0 +1,234 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function up(knex) {
+  return knex.schema.raw(`
+    drop view assignable_campaign_contacts cascade;
+
+    create or replace view assignable_campaign_contacts as (
+      select
+        campaign_contact.id, campaign_contact.campaign_id,
+        campaign_contact.message_status, campaign.texting_hours_end,
+        campaign_contact.timezone::text as contact_timezone
+      from campaign_contact
+      join campaign on campaign_contact.campaign_id = campaign.id
+      where assignment_id is null
+        and auto_reply_eligible = false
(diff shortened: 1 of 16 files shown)


migrations/20230811182109_auto-replies-autoassignment.js
```diff
@@ -0,0 +1,234 @@
...
+      from assignable_campaigns
+      where exists (
+        select 1
+        from assignable_needs_reply
+        where campaign_id = assignable_campaigns.id
+      )
+    );
+
+    drop index todos_partial_idx;
+    create index todos_partial_idx on campaign_contact (campaign_id, assignment_id, message_status, is_opted_out, auto_reply_eligible) where (archived = false);
```
ajohn25, line 115: Highlighting that this adds `auto_reply_eligible` to `todos_partial_idx`. I think this is a good move because the field is functionally pretty similar to `is_opted_out` but welcome any thoughts otherwise!

ajohn25:
This should be ready for review now!

Open Questions:
1. [This test](https://github.com/politics-rewired/Spoke/blob/feat-auto-reply-handling/src/server/api/lib/message-sending.spec.ts#L171) suceeds locally because i have `ENABLE_AUTO_REPLIES` set to true. Curious about thoughts for setting an envvar value / getting it to succeed on CI / CD?
2. Schema dump is failing for me locally, will pair with someone on Monday to figure out why

hiemanshu:
To add to what Chrobot said above, we should have a different config for tests, that enables all the flags possible. We should test with all possible features turned on.

ajohn25:
@bchrobot sorry i converted this back to draft to address a few things i thought of - one of those being that `update-dump` wasn't working for me 😭

PR #1665, opened 2023-08-17
chore(config): check test environment for feature flags

ajohn25:
## Description

This modifies the config file to enable feature flags for running tests.

## Motivation and Context

https://github.com/politics-rewired/Spoke/pull/1663#issuecomment-1681067348

## How Has This Been Tested?

Tests are still passing

## Checklist:

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -23,12 +23,6 @@ const validators = {
     desc: "ActionKit API secret.",
     default: undefined
   }),
-  ALLOW_SEND_ALL: bool({
-    desc:
-      "Whether to allow sending all messages in a campaign at once. NOT LEGAL IN US.",
-    default: false,
-    isClient: true
-  }),
   ALTERNATE_LOGIN_URL: url({
     desc:
       'When set, the "Login" link on the home page will direct to this alternate URL',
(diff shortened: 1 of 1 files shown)


bchrobot (commented):
I think I was not clear enough in my comment https://github.com/politics-rewired/Spoke/pull/1663#issuecomment-1679103314. The envvar definitions have to remain in the `validators` object -- they are only meaningful if passed through `envalid.cleanEnv()`.

Separating the config from the parsed envvars allows something like:

```ts
const config = {
  allowSendAll: env.isTest || env.ALLOW_SEND_ALL
};
```

ajohn25:
> they are only meaningful if passed through envalid.cleanEnv()

Got you! Can you say more about what meaningful refers to here? Is it that the validators aren't really validating anymore because they're not being passed through? The failing test for #1663 passed after [this commit](https://github.com/politics-rewired/Spoke/actions/runs/5894135444/job/15987113776?pr=1663) so i jumped the gun and took that as success 😅

ajohn25:
> then a) TS would have complained that ValidatorSpec<boolean> cannot be assigned to boolean or something like that

Got it! Maybe I'll briefly check whether I can easily convert the Spoke config to TS without breaking anything too 😅

ajohn25:
> I don't think you need to convert to TS to move forward with this. But I think the types are a good way of understanding what is happening?

yeah i agree!

